PR #10, opened 2019-08-12
Add an icon bar, task bar, and border icons all backed by common FDO code

okratitan:
This adds an icon bar/task bar that has a zoom effect.  It also adds icons to border.  All of this code is backed by a common set of utilities for looking up icons, applications, and more using the Freedesktop.org spec.


bar.go
```diff
@@ -1,22 +1,94 @@
...
-		layout.NewSpacer(),
-		widget.NewButton("Browser", func() {
-			exec.Command("chromium").Start()
-		}),
-		widget.NewButton("Terminal", func() {
-			exec.Command("xterm").Start()
-		}),
-		layout.NewSpacer(),
-	)
+var appBar *Bar
```
andydotxyz: as with imports it's best to group these with a single "var (" container

bar.go
```diff
@@ -1,22 +1,94 @@
...
+var fyconSize = 32
+var fyconTheme = "Papirus"
+var icons []*Fycon
+
+type barStackListener struct {
+}
+
+func barCreateFycon(taskbar bool, fdoDesktop *FdoDesktop, win Window) *Fycon {
```
andydotxyz: In Go we avoid duplication so "fdoDesktop *FdoDesktop" is not really best practice.
A descriptive but short name is preferred, such as "data", "source" or "metadata".

bar.go
```diff
@@ -1,22 +1,94 @@
...
+
+func newBar(wm WindowManager) fyne.CanvasObject {
+	appBar = NewAppBar()
+	appBar.Append(layout.NewSpacer())
+
+	if wm != nil {
+		bsl := &barStackListener{}
+		wm.AddStackListener(bsl)
+	}
+	icons := []string{"xterm", "gimp", "google-chrome", "firefox", "xterm", "gimp", "google-chrome", "firefox"}
```
andydotxyz: Can this be extracted as a var at the top of the file as well? it gets lost in the code down here.

barlayout.go
```diff
@@ -0,0 +1,145 @@
...
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+)
+
+// Declare conformity with Layout interface
+var _ fyne.Layout = (*BarLayout)(nil)
+
+//BarLayout returns a layout used for zooming linear groups of icons
+type BarLayout struct {
```
andydotxyz: I don't think this type is needed outside of the bar, so don't export it (i.e. rename to barLayout).
Likewise the fields below are all private so there is no need to export them.

barlayout.go
```diff
@@ -0,0 +1,145 @@
...
+func (bl *BarLayout) SetPointerInside(inside bool) {
+	bl.MouseInside = inside
+}
+
+//SetPointerPosition tells the BarLayout that the mouse position has been updated.
+func (bl *BarLayout) SetPointerPosition(position fyne.Position) {
+	bl.MousePosition = position
+}
+
+func (bl *BarLayout) isSpacer(obj fyne.CanvasObject) bool {
```
andydotxyz: Pretty sure that this and isHorizontalSpacer can be removed in line with the comments about layout below

barlayout.go
```diff
@@ -0,0 +1,145 @@
...
+	// invisible spacers don't impact layout
+	if !obj.Visible() {
+		return false
+	}
+
+	return isHorizontalSpacer(obj)
+}
+
+// Layout is called to pack all child objects into a specified size.
+// For a VBoxLayout this will pack objects into a single column where each item
```
andydotxyz: I was about to say that this looks like a copy of the box layout when I saw this comment ;)
Let's update the docs and strip it back to the required functionality...

barlayout.go
```diff
@@ -0,0 +1,145 @@
...
+// is full width but the height is the minimum required.
+// Any spacers added will pad the view, sharing the space if there are two or more.
+func (bl *BarLayout) Layout(objects []fyne.CanvasObject, size fyne.Size) {
+	spacers := make([]fyne.CanvasObject, 0)
```
andydotxyz: We should not have to manage spacers at all - this is centered in the available space so we can code implicit spacers left and right without needing to include objects to represent them on a list?

barlayout.go
```diff
@@ -0,0 +1,145 @@
...
+// the sum of of all children combined with padding between each.
+func (bl *BarLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
+	spacerCount := 0
+	minSize := fyne.NewSize(0, 0)
+	for _, child := range objects {
+		if !child.Visible() {
+			continue
+		}
+
+		if bl.isSpacer(child) {
```
andydotxyz: As above our minSize is really the size of the icons so our bar can then just center within the space we are told is available.

barwidget.go
```diff
@@ -0,0 +1,178 @@
...
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+)
+
+//Bar is the main widget housing the icon launcher
+type Bar struct {
+	baseWidget
+
+	Children      []fyne.CanvasObject
```
andydotxyz: I don't think any of these need to be exported - does the Bar type either?

barwidget.go
```diff
@@ -0,0 +1,178 @@
...
+	b.show(b)
+}
+
+//Hide makes the widget hidden
+func (b *Bar) Hide() {
+	b.hide(b)
+}
+
+//Prepend adds an object to the begging of the widget
+func (b *Bar) Prepend(object fyne.CanvasObject) {
```
andydotxyz: This feels like a copy paste from the VBox object? I don't think we need to expose any code for handling addition of items to external packages... My guess is that Prepend() is not needed at all and that the Append() below could be private

barwidget.go
```diff
@@ -0,0 +1,178 @@
...
+	if b.Hidden && object.Visible() {
+		object.Hide()
+	}
+	b.Children = append(b.Children, object)
+
+	widget.Refresh(b)
+}
+
+//AppendTaskbar adds an object to the taskbar area of the widget just before the final spacer
+func (b *Bar) AppendTaskbar(object fyne.CanvasObject) {
```
andydotxyz: This seems like a slightly misleading name as it's not appending a taskbar - perhaps "appendToTaskbar" is better?

barwidget.go
```diff
@@ -0,0 +1,178 @@
...
+		object.Hide()
+	}
+	b.Children[len(b.Children)-1] = object
+	b.Append(layout.NewSpacer())
+
+	widget.Refresh(b)
+}
+
+//Remove removes an object from the taskbar area of the widget
+func (b *Bar) Remove(object fyne.CanvasObject) {
```
andydotxyz: This name should be consistent with the above so that it's clearly also taskbar related

barwidget.go
```diff
@@ -0,0 +1,178 @@
...
+	Bar := &Bar{Children: children}
+	widget.Renderer(Bar).Layout(Bar.MinSize())
+	return Bar
+}
+
+//BarRenderer privdes the renderer functions for the Bar Widget
+type BarRenderer struct {
+	layout BarLayout
+
+	Bar     *Bar
```
andydotxyz: This should be private, as should it's renderer parent

basewidget.go
```diff
@@ -0,0 +1,88 @@
+package desktop
+
+import (
```
andydotxyz: As this is a copy of the baseWidget in fyne can you write that in a comment along with an indicator that we should remove it once the new, better, BaseWidget is complete? linking to https://github.com/fyne-io/fyne/issues/324

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+	"bufio"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"fyne.io/fyne"
+)
+
+var iconExtensions = []string{".png", ".svg", ".xpm"}
```
andydotxyz: I don't think we should search .xpm as we cannot render it..?

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+
+//FdoDesktop contains the information from Linux .desktop files
+type FdoDesktop struct {
```
andydotxyz: This does not seem to be Fdo specific, and we should not really export a namespaced type name.
We could search and replace this as "IconData" type or create an IconData interface that this type privately implements.
This is because we don't want to pass any FDO naming outside of the implementation detail for the type.

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+	Name     string
+	IconName string
+	IconPath string
+	Exec     string
+}
+
+//FdoResourceFormat turns a string into a usable name for fyne.Resources
+func FdoResourceFormat(name string) string {
```
andydotxyz: Perhaps this is "ResourceName" as it's not really a format.
Aditionally would it not be better to encapsulate this privately and expose a "LoadIcon() fyne.Resource" method to avoid duplicated use of this internal detail?

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+		var fallbackLocations []string
+		fallbackLocations = append(fallbackLocations, "/usr/local/share")
+		fallbackLocations = append(fallbackLocations, "/usr/share")
+		return fallbackLocations
+	}
+	return locationLookup
+}
+
+//FdoLookupApplication looks up an application by name and returns a FdoDesktop
+func FdoLookupApplication(appName string) *FdoDesktop {
```
andydotxyz: keep this private too.

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+	desktop = FdoLookupApplication(iconName)
+	return desktop
+}
+
+func reverseWalkDirectoryMatch(directory string, joiner string, iconName string) string {
+	//If the requested icon wasn't found in the Example 32x32 or scalable dirs of the theme, try other sizes within theme
+	files, err := ioutil.ReadDir(directory)
+	if err != nil {
+		fyne.LogError("", err)
```
andydotxyz: Can we provide an error string here (even just "Icon directory read error")? makes debugging easier

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+		return ""
+	}
+	//Lets walk the files in reverse so bigger icons are selected first (Unless it is a 3 digit icon size like 128)
+	for i := len(files) - 1; i >= 0; i-- {
+		f := files[i]
+		if strings.HasPrefix(f.Name(), ".") == false {
```
andydotxyz: This can also be less complex by skipping early...

```go
if strings.HasPrefix(f.Name(), ".") || !f.IsDir() {
    continue
}
```

leaves the rest of the method less indented

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+				matchDir := filepath.Join(directory, f.Name())
+				for _, extension := range iconExtensions {
+					//Example is /usr/share/icons/icon_theme/64x64/apps/xterm.png
+					var testIcon = ""
+					if joiner != "" {
+						testIcon = filepath.Join(matchDir, joiner, iconName+extension)
+					} else {
+						testIcon = filepath.Join(matchDir, iconName+extension)
+					}
+					if _, err := os.Stat(testIcon); os.IsNotExist(err) {
```
andydotxyz: can this be inverted?
```go
if _, err := os.Stat(testIcon); err == nil {
    return testIcon
}
```
no need for the continue block then

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+}
+
+func lookupIconPathInTheme(dir string, iconName string) string {
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return ""
+	}
+	for _, extension := range iconExtensions {
+		//Example is /usr/share/icons/icon_theme/32x32/apps/xterm.png
+		testIcon := filepath.Join(dir, string(fyconSize), "apps", iconName+extension)
+		if _, err := os.Stat(testIcon); os.IsNotExist(err) {
```
andydotxyz: This is lots of repeated code each indenting further. As immediately above, reverse the logic then you can return in the if block - otherwise move on to the next line...?

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+				return testIcon
+			}
+		} else {
+			return testIcon
+		}
+	}
+	for _, extension := range iconExtensions {
+		//Example is /usr/share/icons/icon_theme/scalable/apps/xterm.png - Try this if the requested size didn't exist
+		testIcon := filepath.Join(dir, "scalable", "apps", iconName+extension)
+		if _, err := os.Stat(testIcon); os.IsNotExist(err) {
```
andydotxyz: As above, invert logic to simplify

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+	for _, dataDir := range locationLookup {
+		//Icon was not found in default theme or default fallback theme - Check all themes for a match
+		//Example is /usr/share/icons
+		files, err := ioutil.ReadDir(dataDir + "icons")
+		if err != nil {
+			fyne.LogError("", err)
+			continue
+		}
+		//Enter icon theme
+		for _, f := range files {
```
andydotxyz: This whole block gets very complex. I think inverting the logic, as above, should help simplify it.
Less indentation is a good indication of lower complexity.

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+			if strings.HasPrefix(f.Name(), ".") == false {
+				if f.IsDir() {
+					//Example is /usr/share/icons/gnome
+
+					subFiles, err := ioutil.ReadDir(filepath.Join(dataDir, "icons", f.Name()))
+					if err != nil {
+						fyne.LogError("", err)
```
andydotxyz: Please include an error reason

fdo.go
```diff
@@ -0,0 +1,264 @@
...
+			fdoDesktop.IconPath = FdoLookupIconPath(fdoDesktop.IconName)
+		} else if strings.HasPrefix(line, "Exec=") {
+			exec := strings.SplitAfter(line, "=")
+			fdoDesktop.Exec = exec[1]
+		} else {
+			continue
+		}
+	}
+	if err := scanner.Err(); err != nil {
+		fyne.LogError("", err)
```
andydotxyz: This needs an error reason too :)

fycon.go
```diff
@@ -0,0 +1,127 @@
...
+
+func (fy *fyconRenderer) Destroy() {
+}
+
+// Fycon widget is a basic image component that load's its resource to match the theme.
+type Fycon struct {
+	baseWidget
+
+	OnTapped      func()
+	Resource      fyne.Resource // The resource for this Fycon
```
andydotxyz: I don't think that this needs to be public - and the comment does not add anything here - something a little more meaningful - and include for the one above and below

fycon.go
```diff
@@ -0,0 +1,127 @@
...
+func (fy *Fycon) Show() {
+	fy.show(fy)
+}
+
+// Hide this widget, if it was previously visible
+func (fy *Fycon) Hide() {
+	fy.hide(fy)
+}
+
+// SetResource updates the resource rendered in this Fycon widget
```
andydotxyz: This function is not used anywhere - do we need it?

wm/border.go
```diff
@@ -1,19 +1,42 @@
...
 	"fyne.io/fyne"
+
 	"fyne.io/fyne/canvas"
 	"fyne.io/fyne/layout"
 	"fyne.io/fyne/theme"
 	"fyne.io/fyne/widget"
 )
 
-func newBorder(title string) fyne.CanvasObject {
+func newBorder(title string, class []string, command string, iconName string) fyne.CanvasObject {
```
andydotxyz: To save exploding this list of parameters can we not pass a desktop.Window?

wm/property.go
```diff
@@ -22,6 +23,44 @@ func windowName(x *xgbutil.XUtil, win xproto.Window) string {
 	return name
 }
 
+func windowClass(x *xgbutil.XUtil, win xproto.Window) []string {
+	class, err := xprop.PropValStrs(xprop.GetProperty(x, win, "WM_CLASS"))
+	if err != nil {
+		class, err := xprop.PropValStrs(xprop.GetProperty(x, win, "_NET_WM_CLASS"))
+		if err != nil {
+			return []string{"Noclass"}
```
andydotxyz: Surely in these 3 cases a zero value ("") would be better than an arbitrary string?
If this string is needed then it should probably be a public const so that others can compare to it...

andydotxyz:
Oh, as well as the code review notes I see that there are no tests, so this will reduce the overall test coverage which will fail the build as soon as it's merged in - can this be addressed?

bar.go
```diff
@@ -1,22 +1,94 @@
...
-			exec.Command("chromium").Start()
-		}),
-		widget.NewButton("Terminal", func() {
-			exec.Command("xterm").Start()
-		}),
-		layout.NewSpacer(),
-	)
+var appBar *Bar
+var fyconSize = 32
+var fyconTheme = "Papirus"
```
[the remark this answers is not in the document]
okratitan: Hicolor is already hard coded in as the fallback - but I can change this to hicolor for now... It doesn't matter as much because I plan to add a settings panel for this as soon as this pr is accepted.

fdo.go
okratitan: It is Fdo specific as it is all functions related to fdo spec.  Also this currently has to be exported for use by the window border to get the border icon and that is in a different package from desktop

fdo.go
okratitan: There are issues with keeping this private as I mentioned above

wm/border.go
okratitan: No - I don't think I have a reference to a window at the time newBorder is being called in wm/frame.go?

fdo.go
okratitan: I took a stab at doing an interface, if it sucks I will gladly revert and we can move forward as planned.

bar.go
```diff
@@ -54,8 +55,8 @@ func barCreateIcon(taskbar bool, idata *IconData, win Window) *barIcon {
 }
 
 func (bsl *barStackListener) WindowAdded(win Window) {
-	fdoDesktop := FdoLookupApplicationWinInfo(win.Title(), win.Class(), win.Command(), win.IconName())
-	icon := barCreateIcon(true, fdoDesktop, win)
+	idata := driver.GetIconDataByWinInfo(iconTheme, iconSize, win.Title(), win.Class(), win.Command(), win.IconName())
```
andydotxyz: I'm not sure about the "i" prefix here... is it for interface? if so we don't do that in Go.
If it's part of the name then maybe "data" or "iconData" is better.

driver/fdo.go
```diff
@@ -11,18 +11,54 @@ import (
 	"fyne.io/fyne"
 )
 
-var iconExtensions = []string{".png", ".svg"}
+var (
+	iconExtensions = []string{".png", ".svg"}
+	iconTheme      string
```
andydotxyz: Having these global variables in the driver package seems strange (I'll reference them below)

driver/fdo.go
```diff
@@ -11,18 +11,54 @@ import (
...
+type fdoIconData struct {
+	name     string // Application name
+	iconName string // Icon name
+	iconPath string // Icon path
+	exec     string // Command to execute application
+}
+
+//Returns the name associated with an fdo app
+func (idata *fdoIconData) Name() string {
+	if idata == nil {
```
andydotxyz: Checking for nil should not be needed here, it's reasonable to assume that our "self" type is not nil.

driver/fdo.go
```diff
@@ -34,35 +70,39 @@ func FdoLookupXdgDataDirs() []string {
 	return locationLookup
 }
 
-//FdoLookupApplication looks up an application by name and returns an IconData struct
-func FdoLookupApplication(appName string) *IconData {
-	locationLookup := FdoLookupXdgDataDirs()
+//fdoLookupApplication looks up an application by name and returns an fdoIconData struct
+func fdoLookupApplication(theme string, size int, appName string) *fdoIconData {
+	iconTheme = theme
```
andydotxyz: Setting globals as a side effect of a "read only" function (such as get, lookup etc) is a bit of a smell.
If this function is called for different themes the results could be undefined... surely these global values are not needed?

bar.go
```diff
@@ -22,23 +23,23 @@ var (
 type barStackListener struct {
 }
 
-func barCreateIcon(taskbar bool, idata *IconData, win Window) *barIcon {
-	if idata == nil || idata.IconPath == "" {
+func barCreateIcon(taskbar bool, idata driver.IconData, win Window) *barIcon {
```
andydotxyz: Having the top level code call into the driver is a backwards dependency - I think the IconData type is in the wrong package.

wm/border.go
andydotxyz: Yes we do. newBorder is called from *frame, which is itself a Window...

driver/icons.go
```diff
@@ -0,0 +1,21 @@
+package driver
```
andydotxyz: This interface is just the thing, but interfaces should be at the top level, where sub-packages are free to implement them without the main code having to know where to look

andydotxyz:
Thanks, the general idea is right, details in the review.
However the tests are failing and also the PR notes that landing this will drop the test coverage, indicating that this code needs more tests.

PR #12, opened 2019-08-20
This improves the results of fdo searching and adds tests to the fdo methods

okratitan:
This improves the results of fdo searching by also searching icons in sub folders besides apps, comparing our lookup strings to the contents of desktop files if filename matching doesn't work, looks in pixmaps for icon files, and skips the lookup if the icon name is a path to an icon.

The test added creates a test xdg/fdo environment and should cover most if not all cases for icon lookups in fdo.go


internal/fdo.go
```diff
@@ -64,9 +64,57 @@ func fdoLookupApplication(theme string, size int, appName string) desktop.IconDa
 			return newFdoIconData(theme, size, testLocation)
 		}
 	}
+	//No Match Was Found Checking File Names - Check Files Themselves
```
andydotxyz: This comment is slightly ambiguous. You mean "file contents" right?
When a comment has to be written it indicates a new function with a clear name could be used.

How about "fdoLookupApplicationByMetadata" to fit with your current naming?

internal/fdo.go
```diff
@@ -64,9 +64,57 @@ func fdoLookupApplication(theme string, size int, appName string) desktop.IconDa
...
+					return icon
+				}
+			}
+		}
+	}
 	return nil
 }
 
+//fdoLookupApplicationPartial looks up an application by a partial name and returns all matches in an fdoIconData struct slice
+func fdoLookupApplicationPartialMatch(theme string, size int, appName string) []desktop.IconData {
```
andydotxyz: This should be plural as it returns many applications.
I wonder if just "Matching" rather than "PartialMatch" is clearer?

internal/fdo.go
```diff
@@ -64,9 +64,57 @@ func fdoLookupApplication(theme string, size int, appName string) desktop.IconDa
...
+	var icons []desktop.IconData
+	locationLookup := fdoLookupXdgDataDirs()
+	for _, dataDir := range locationLookup {
```
andydotxyz: This chunk of code is almost a compete duplicate of the section above.
A helper function such as "runOnEachApplicationFile"? this would make these functions shorter, reduce the duplication and make it clearer of the purpose.

internal/fdo.go
```diff
@@ -232,23 +297,33 @@ func newFdoIconData(theme string, size int, desktopPath string) *fdoIconData {
...
-		} else if strings.HasPrefix(line, "Icon=") {
-			icon := strings.SplitAfter(line, "=")
-			fdoApp.iconName = icon[1]
-			fdoApp.iconPath = fdoLookupIconPath(theme, size, fdoApp.iconName)
-			if fdoApp.iconPath == "" {
-				fyne.LogError("Could not find path for icon "+fdoApp.iconName, nil)
+		if strings.HasPrefix(line, "[") {
+			currentSection = line
+		}
+		if currentSection == "[Desktop Entry]" {
```
andydotxyz: invert the logic here again and you can avoid this additional indentation of the main logic

internal/fdo.go
```diff
@@ -232,23 +297,33 @@ func newFdoIconData(theme string, size int, desktopPath string) *fdoIconData {
...
+					fdoApp.iconPath = fdoLookupIconPath(theme, size, fdoApp.iconName)
+				}
+				if fdoApp.iconPath == "" {
+					fyne.LogError("Could not find path for icon "+fdoApp.iconName, nil)
+				}
+			} else if strings.HasPrefix(line, "Exec=") {
+				exec := strings.SplitAfter(line, "=")
+				fdoApp.exec = exec[1]
+			} else {
+				continue
```
andydotxyz: I don't think you need a continue at the end of this

internal/fdo_test.go
```diff
@@ -0,0 +1,204 @@
...
+	"github.com/magiconair/properties/assert"
+)
+
+var iconTheme = "default_theme"
+var iconSize = 32
+
+type dummyWindow struct {
+}
+
+type dummyWindow1 struct {
```
andydotxyz: This is a lot of boilerplate structs - why not just create a single one with some state?
It would be much easier to read - you'd put the setup code next to the code testing it.

internal/fdo_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+//applications/app3.desktop and applications/app3.png
+func TestFdoIconNameIsPath(t *testing.T) {
+	setTestEnv()
+	dataLocation := os.Getenv("XDG_DATA_DIRS")
+	output := fmt.Sprintf("[Desktop Entry]\nName=App3\nExec=app3\nIcon=%s\n", filepath.Join(dataLocation, "icons", "app3.png"))
+	fmt.Print(output)
+	err := ioutil.WriteFile(filepath.Join(dataLocation, "applications", "app3.desktop"), []byte(output), 0644)
+	if err != nil {
+		fyne.LogError("Could not create desktop for Icon Name path example", err)
```
andydotxyz: In unit tests it is better to fail than to log an error

internal/fdo_test.go
```diff
@@ -0,0 +1,204 @@
...
+			test = true
+		}
+	}
+	return test
+}
+
+func setTestEnv() {
+	workingDir, err := os.Getwd()
+	if err != nil {
+		fyne.LogError("Could not get current working directory", err)
```
andydotxyz: Ignoring this error (or logging it) is bad - the test code around line 125 will then write files into the root of the filesystem

internal/fdo_test.go
```diff
@@ -0,0 +1,204 @@
...
+func TestFdoLookupIconNotInApps(t *testing.T) {
+	setTestEnv()
+	data := fdoLookupApplication(iconTheme, iconSize, "app9")
+	assert.Equal(t, exists(data), true)
+}
+
+func TestFdoLookupIconByWinInfo(t *testing.T) {
+	setTestEnv()
+	//Test win info lookup by title
+	win1 := &dummyWindow1{}
```
andydotxyz: As above the multiple structs make it hard to read this - "dummyWindow1" could be "dummyWIndow{title="blah"}" - then we have test setup next to test code

internal/testdata/applications/app3.desktop
```diff
@@ -0,0 +1,4 @@
+[Desktop Entry]
+Name=App3
+Exec=app3
+Icon=/home/stephen/go/src/fyne.io/desktop/internal/testdata/icons/app3.png
```
andydotxyz: This doesn't seem right?

internal/fdo.go
okratitan: Hmmm... I think on the surface they seem like almost a complete duplicate - but in reality they aren't at all.  The section above stops the loop as soon as it finds an icon and it returns a single icon.  It also looks for exact matches.  This loop does not stop the loop ever and appends every match that is non exact to a slice and then returns.  So in reality one is runOnEachApplicationFileUntilMatched and the other is runOnEachApplicationFile.

Because the loops don't behave the same - one exits early the other must continue - and because the logic within the loops are different - one checks for only one exact match, and the other for multiple partial matches, and that the return values are of different types as well... I'm finding it hard to think of a clean way to separate.

PR #16, opened 2019-08-20
Add a nice App Exec Button with Live Preview

okratitan:
This adds a nice app exec button to the widget panel and when clicked opens a live preview that populates all matching apps as you type so you can easily select an application to open.


andydotxyz:
Should this not use a window rather than an overlay?
The overlay will appear over all the root window content but will not appear above any other windows currently loaded...

PR #17, opened 2019-08-26
Client refactor

okratitan:
This refactors the wm to have another layer: client. It exists between the wm and the frame to manage windows, their hints, messaging, and properties. This also adds a minimize and maximize button with according functionality to the border. Finally, this fixes the appBar taskbar to work as a proper taskbar with the new features.

### DIFF
--- a/bar.go
+++ b/bar.go
@@ -43,6 +43,14 @@ func (b *bar) WindowAdded(win Window) {
 	icon := barCreateIcon(b, true, data, win)
 	if icon != nil {
 		icon.onTapped = func() {
+			if !win.Iconic() && win.TopWindow() {
+				win.Iconify()
+				return
+			}
+			if win.Iconic() {
+				win.Uniconify()
+			}
+			win.RaiseToTop()
 			win.Focus()
 		}
 		appBar.append(icon)
@@ -53,8 +61,10 @@ func (b *bar) WindowRemoved(win Window) {
 	for i, icon := range icons {
 		if icon.taskbarWindow != nil {
 			if win == icon.taskbarWindow {
-				appBar.removeFromTaskbar(icon)
-				icons = append(icons[:i], icons[i+1:]...)
+				if !win.Iconic() {
+					appBar.removeFromTaskbar(icon)
+					icons = append(icons[:i], icons[i+1:]...)
+				}
 			}
 		}
 	}
--- a/bar.go
+++ b/bar.go
@@ -35,6 +35,18 @@ func barCreateIcon(b *bar, taskbar bool, data AppData, win Window) *barIcon {
 	return icon
 }
 
+func taskbarIconTapped(win Window) {
+	if !win.Iconic() && win.TopWindow() {
+		win.Iconify()
+		return
+	}
+	if win.Iconic() {
+		win.Uniconify()
+	}
+	win.RaiseToTop()
+	win.Focus()
+}
+
 func (b *bar) WindowAdded(win Window) {
 	data := b.desk.IconProvider().FindAppFromWinInfo(win)
 	if data == nil {
@@ -43,15 +55,7 @@ func (b *bar) WindowAdded(win Window) {
 	icon := barCreateIcon(b, true, data, win)
 	if icon != nil {
 		icon.onTapped = func() {
-			if !win.Iconic() && win.TopWindow() {
-				win.Iconify()
-				return
-			}
-			if win.Iconic() {
-				win.Uniconify()
-			}
-			win.RaiseToTop()
-			win.Focus()
+			taskbarIconTapped(win)
 		}
 		appBar.append(icon)
 	}
@@ -59,13 +63,12 @@ func (b *bar) WindowAdded(win Window) {
 
 func (b *bar) WindowRemoved(win Window) {
 	for i, icon := range icons {
-		if icon.taskbarWindow != nil {
-			if win == icon.taskbarWindow {
-				if !win.Iconic() {
-					appBar.removeFromTaskbar(icon)
-					icons = append(icons[:i], icons[i+1:]...)
-				}
-			}
+		if icon.taskbarWindow == nil || win != icon.taskbarWindow {
+			continue
+		}
+		if !win.Iconic() {
+			appBar.removeFromTaskbar(icon)
+			icons = append(icons[:i], icons[i+1:]...)
 		}
 	}
 }
--- a/barwidget_test.go
+++ b/barwidget_test.go
@@ -1,9 +1,10 @@
 package desktop
 
 import (
+	"testing"
+
 	"fyne.io/fyne"
 	"github.com/stretchr/testify/assert"
-	"testing"
 
 	_ "fyne.io/fyne/test"
 	"fyne.io/fyne/widget"
@@ -31,6 +32,17 @@ func (*dummyWindow) Command() string {
 func (*dummyWindow) IconName() string {
 	return "xterm"
 }
+func (*dummyWindow) Iconic() bool {
+	return false
+}
+
+func (*dummyWindow) Maximized() bool {
+	return false
+}
+
+func (*dummyWindow) TopWindow() bool {
+	return true
+}
 
 func (*dummyWindow) Focus() {
 	// no-op
@@ -40,9 +52,30 @@ func (*dummyWindow) Close() {
 	// no-op
 }
 
-func (*dummyWindow) RaiseAbove(Window) {
+func (*dummyWindow) Iconify() {
+	// no-op
+}
+
+func (*dummyWindow) Uniconify() {
+	// no-op
+}
+
+func (*dummyWindow) Maximize() {
+	// no-op
+}
+
+func (*dummyWindow) Unmaximize() {
+	// no-op
+}
+
+func (*dummyWindow) RaiseAbove(desktop.Window) {
 	// no-op (this is instructing the window after stack changes)
 }
+
+func (*dummyWindow) RaiseToTop() {
+	// no-op
+}
+
 
 type dummyIcon struct {
 }
--- a/barwidget_test.go
+++ b/barwidget_test.go
@@ -68,14 +68,13 @@ func (*dummyWindow) Unmaximize() {
 	// no-op
 }
 
-func (*dummyWindow) RaiseAbove(desktop.Window) {
+func (*dummyWindow) RaiseAbove(Window) {
 	// no-op (this is instructing the window after stack changes)
 }
 
 func (*dummyWindow) RaiseToTop() {
 	// no-op
 }
-
 
 type dummyIcon struct {
 }
--- a/theme/theme.go
+++ b/theme/theme.go
@@ -15,6 +15,10 @@ var (
 	BrightnessIcon = theme.NewThemedResource(brightnessIcon, nil)
 	// BrokenImageIcon is the material design icon for a broken image
 	BrokenImageIcon = theme.NewThemedResource(brokenImageIcon, nil)
+	// MaximizeIcon is the material design icon for maximizing a window
+	MaximizeIcon    = theme.NewThemedResource(maximizeIcon, nil)
+	// MinimizeIcon is the material design icon for minimizing a window
+	MinimizeIcon = theme.NewThemedResource(minimizeIcon, nil)
 
 	// BorderWidth is the width of window frames
 	BorderWidth = 4
--- a/theme/theme.go
+++ b/theme/theme.go
@@ -16,9 +16,9 @@ var (
 	// BrokenImageIcon is the material design icon for a broken image
 	BrokenImageIcon = theme.NewThemedResource(brokenImageIcon, nil)
 	// MaximizeIcon is the material design icon for maximizing a window
-	MaximizeIcon    = theme.NewThemedResource(maximizeIcon, nil)
+	MaximizeIcon = theme.NewThemedResource(maximizeIcon, nil)
 	// MinimizeIcon is the material design icon for minimizing a window
-	MinimizeIcon = theme.NewThemedResource(minimizeIcon, nil)
+	IconifyIcon = theme.NewThemedResource(iconifyIcon, nil)
 
 	// BorderWidth is the width of window frames
 	BorderWidth = 4
--- a/wm/client.go
+++ b/wm/client.go
@@ -27,11 +27,8 @@ type client struct {
 	iconic    bool
 	maximized bool
 
-	allowedActions []string
-	stateHints     []string
-
-	f  *frame
-	wm *x11WM
+	frame *frame
+	wm    *x11WM
 }
 
 func (s *stack) clientForWin(id xproto.Window) desktop.Window {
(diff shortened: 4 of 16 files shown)


bar.go
```diff
@@ -43,6 +43,14 @@ func (b *bar) WindowAdded(win Window) {
 	icon := barCreateIcon(b, true, data, win)
 	if icon != nil {
 		icon.onTapped = func() {
+			if !win.Iconic() && win.TopWindow() {
```
andydotxyz: This is getting complex, it should probably be a separate function (named, not inline).
Also why does it only iconify top level windows?

bar.go
```diff
@@ -53,8 +61,10 @@ func (b *bar) WindowRemoved(win Window) {
 	for i, icon := range icons {
 		if icon.taskbarWindow != nil {
```
andydotxyz: This is getting complex again. Let's invert the logic, i.e.

`if icon.taskbarWindow == nil || win != icon.taskbarWindow { continue }`

barwidget_test.go
```diff
@@ -40,10 +52,31 @@ func (*dummyWindow) Close() {
...
+
+func (*dummyWindow) Maximize() {
+	// no-op
+}
+
+func (*dummyWindow) Unmaximize() {
+	// no-op
+}
+
+func (*dummyWindow) RaiseAbove(desktop.Window) {
```
andydotxyz: the desktop prefix should not be here surely?

theme/theme.go
```diff
@@ -15,6 +15,10 @@ var (
 	BrightnessIcon = theme.NewThemedResource(brightnessIcon, nil)
 	// BrokenImageIcon is the material design icon for a broken image
 	BrokenImageIcon = theme.NewThemedResource(brokenImageIcon, nil)
+	// MaximizeIcon is the material design icon for maximizing a window
+	MaximizeIcon    = theme.NewThemedResource(maximizeIcon, nil)
+	// MinimizeIcon is the material design icon for minimizing a window
+	MinimizeIcon = theme.NewThemedResource(minimizeIcon, nil)
```
andydotxyz: In the API this is referred to as "Iconic" but here "Minimize" - we should be consistent...

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+
+	return nil
+}
+
+func (cli *client) Decorated() bool {
+	return cli.framed
+}
+
+func (cli *client) Title() string {
+	if cli.title == "" {
```
andydotxyz: Why are these being cached? What happens if the hint changes?
It seems like a premature optimisation...

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+	iconName  string
+
+	allowedActions []string
+	stateHints     []string
+
+	fr *frame
+	wm *x11WM
+}
+
+func (s *stack) clientForWin(id xproto.Window) desktop.Window {
```
andydotxyz: clientForWin returns a Window - seems like a mismatch?

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+	for _, w := range s.clients {
+		if w.(*client).id == id || w.(*client).win == id {
+			return w
+		}
+	}
+
+	return nil
+}
+
+func (cli *client) Decorated() bool {
```
andydotxyz: Using a shortened name like "cli" is not common - perhaps just "c" as the context is clear... or "client" as it's a word.

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+	cm, err := xevent.NewClientMessage(32, cli.win, protocols,
+		int(delWin))
+	err = xproto.SendEventChecked(cli.wm.x.Conn(), false, cli.win, 0,
+		string(cm.Bytes())).Check()
+	if err != nil {
+		fyne.LogError("Window Delete Error", err)
+	}
+}
+
+func (cli *client) iconifyMessage(state int) {
```
andydotxyz: This is not really iconifyMessage - it is actually "sendMessage".
the "iconifyAtm" should just be "stateChangeAtom" and it can be used by other parts of code.

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+func (cli *client) maximizeMessage(action int) {
+	ewmh.WmStateReqExtra(cli.wm.x, cli.win, action, "_NET_WM_STATE_MAXIMIZED_VERT",
+		"_NET_WM_STATE_MAXIMIZED_HORZ", 1)
+}
+
+func (cli *client) Maximize() {
+	cli.maximized = true
+	//1 is for adding a new state
```
andydotxyz: comment = bad. Better would be to define an enum with ExtraStateEnable and ExtraStateDisable

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+	cli.iconifyMessage(icccm.StateIconic)
+	cli.iconic = true
+}
+
+func (cli *client) Uniconify() {
+	cli.iconifyMessage(icccm.StateNormal)
+	cli.iconic = false
+}
+
+func (cli *client) maximizeMessage(action int) {
```
andydotxyz: As mentioned below the "int" action is overly broad.
Also should the same action not be passed to VERT and HORZ?

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+	for i, curHint := range cli.stateHints {
+		if curHint == hint {
+			cli.stateHints = append(cli.stateHints[:i], cli.stateHints[i+1:]...)
+			ewmh.WmStateSet(cli.wm.x, cli.win, cli.stateHints)
+			return
+		}
+	}
+}
+
+func (cli *client) maximizeFrame() {
```
andydotxyz: These two methods are just proxying to the frame, but we are in the same package - it seems unneeded

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+	maximized bool
+	title     string
+	class     []string
+	command   string
+	iconName  string
+
+	allowedActions []string
+	stateHints     []string
+
+	fr *frame
```
andydotxyz: "fr"? :)

wm/desk.go
```diff
@@ -192,14 +199,125 @@ func (x *x11WM) configureWindow(win xproto.Window, ev xproto.ConfigureRequestEve
 	}
 }
 
+func (x *x11WM) clientMessage(ev xproto.ClientMessageEvent) {
```
andydotxyz: This is a big method with many indented switch statements that should probably be broken out to separate functions. Also "clientMessage" sounds like an action but it's really processing an event no?

wm/desk.go
```diff
@@ -192,14 +199,125 @@ func (x *x11WM) configureWindow(win xproto.Window, ev xproto.ConfigureRequestEve
...
+		}
+	case "_NET_WM_STATE":
+		subMsgAtom, err := xprop.AtomName(x.x, xproto.Atom(ev.Data.Data32[1]))
+		if err != nil {
+			fyne.LogError("Error getting event", err)
+			return
+		}
+		switch subMsgAtom {
+		case "_NET_WM_STATE_HIDDEN":
+			switch ev.Data.Data32[0] {
```
andydotxyz: This seems like a repeating pattern - can it be factored out?

wm/desk.go
```diff
@@ -192,14 +199,125 @@ func (x *x11WM) configureWindow(win xproto.Window, ev xproto.ConfigureRequestEve
...
+			case 0: // Remove Atom
```
andydotxyz: Perhaps consts again? AtomStateRemove = 0, AtomStateAdd = 1...

wm/desk.go
```diff
@@ -192,14 +199,125 @@ func (x *x11WM) configureWindow(win xproto.Window, ev xproto.ConfigureRequestEve
...
+					} else {
+						x.maximizeWindow(ev.Window)
+					}
+				}
+			}
+		}
+	}
+}
+
+func (x *x11WM) iconifyWindow(win xproto.Window) {
```
andydotxyz: It seems like these belong in property.go along with the other hint code?

wm/frame.go
```diff
@@ -3,271 +3,214 @@
...
-
-	if !askNicely {
-		err := xproto.DestroyWindowChecked(f.wm.x.Conn(), f.win).Check()
-		if err != nil {
-			log.Println("Close Err", err)
-		}
-
-		return
-	}
+	cli := fr.cli.wm.clientForWin(fr.cli.win)
```
andydotxyz: This seems strange - isn't that just getting cli from cli's window id???

wm/frame.go
```diff
@@ -3,271 +3,214 @@
...
-func (f *frame) unFrame() {
-	if !f.framed {
-		return
-	}
-
-	frame := f.wm.frameForWin(f.win)
-	f.wm.RemoveWindow(frame)
-
-	if frame == nil {
+func (fr *frame) unFrame() {
```
andydotxyz: Why the change from "f" to "fr"? it's not common and is making the diff bigger than it needs to be.

wm/frame.go
```diff
@@ -3,271 +3,214 @@
...
+	fr.cli.wm.RaiseToTop(fr.cli)
 }
 
-func (f *frame) release(x, y int16) {
-	if x > int16(f.wm.borderWidth()) && x < int16(f.wm.borderWidth()+f.wm.buttonWidth()) &&
-		y < int16(f.wm.borderWidth()+f.wm.titleHeight()) {
-		f.Close()
+func (fr *frame) release(x, y int16) {
+	if x > int16(fr.cli.wm.borderWidth()) && x < int16(fr.cli.wm.borderWidth()+fr.cli.wm.buttonWidth()) &&
+		y < int16(fr.cli.wm.borderWidth()+fr.cli.wm.titleHeight()) {
```
andydotxyz: The y coordinate is the same for all of these - check it at the top and this gets less complex

wm/frame.go
```diff
@@ -3,271 +3,214 @@
...
+		fyne.LogError("Set Cursor Error", err)
 	}
 }
 
-func (f *frame) RaiseAbove(win desktop.Window) {
-	topID := f.wm.rootID
-	if win != nil {
-		topID = win.(*frame).id
+func (fr *frame) updateGeometry(x uint16, y uint16, w uint16, h uint16, moveOnly bool) {
+	if fr.framed && !moveOnly {
```
andydotxyz: if this is not true shouldn't we just return? There is no need to re-configure with unchanged values

wm/frame.go
```diff
@@ -3,271 +3,214 @@
...
+	err := xproto.ConfigureWindowChecked(fr.cli.wm.x.Conn(), fr.cli.id, xproto.ConfigWindowX|xproto.ConfigWindowY|
+		xproto.ConfigWindowWidth|xproto.ConfigWindowHeight,
+		[]uint32{uint32(x), uint32(y), uint32(w), uint32(h)}).Check()
+	if err != nil {
+		fyne.LogError("Configure Window Error", err)
 	}
+}
 
-	f.wm.raiseWinAboveID(f.id, topID)
+func (fr *frame) maximizeFrame() {
```
andydotxyz: No need to repeat naming here - we know it's a frame

wm/frame.go
```diff
@@ -3,271 +3,214 @@
...
+	var w = fr.cli.wm.x.Screen().WidthInPixels
+	var h = fr.cli.wm.x.Screen().HeightInPixels
+	var x, y = 0, 0
```
andydotxyz: These don't seem to be set - is this a bug?

wm/frame.go
```diff
@@ -3,271 +3,214 @@
...
-		log.Println("SetCursor Err", err)
+		fyne.LogError("Set Cursor Error", err)
 	}
 }
 
-func (f *frame) RaiseAbove(win desktop.Window) {
-	topID := f.wm.rootID
-	if win != nil {
-		topID = win.(*frame).id
+func (fr *frame) updateGeometry(x uint16, y uint16, w uint16, h uint16, moveOnly bool) {
```
andydotxyz: This function is always passed the frame x, y, width, height. Either this function should update the values - or it should not take the additional parameters...

wm/client.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/BurntSushi/xgbutil/xprop"
+
+	"fyne.io/desktop"
+	"fyne.io/fyne"
+)
+
+type client struct {
+	id, win xproto.Window
+
+	framed    bool
```
andydotxyz: This seems to mirror or override the one in frame - surely one is not needed?

bar.go
andydotxyz: Oh yes, I thought this was the iconify button on the frame, my bad

wm/desk.go
andydotxyz: I don't agree that code like this is ever a necessary evil...

Something like go```
func handleClientAtom(request int, enable, disable func(), enabled bool) {
...
```

that could be re-used for any instance of this code by passing `(ev.Data.Data32[0], func() {doenable}, func(){do disable}, window.toggleCheck())`

File: wm/desk.go (hunk 16 above)
Comment (andydotxyz) [reply wm/desk.go]: the file is called properties.go, not propertygetters.go :-p

File: wm/frame.go (hunk 17 above)
Comment (andydotxyz) [reply wm/frame.go]: Then why is it called `clientForWin`?

File: wm/frame.go (hunk 22 above)
Comment (andydotxyz) [reply wm/frame.go]: top left of a screen is always 0, 0?

Comment (andydotxyz) [reply wm/frame.go]: Ah yes we need to access the xinerama extension to handle multiple. ignoring for now seems ok

File: wm/client.go (hunk 6 above)
Comment (andydotxyz) [reply wm/client.go]: I think that was because of testable/non-testable file boundaries, but this is now inside client.go, so returning *client might be fine in this instance?

File: wm/frame.go (hunk 17 above)
Comment (andydotxyz) [reply wm/frame.go]: what I mean is, looking at the code is `cli` not the same as `fr.cli`?
If they are different this is a naming issue if it's the same then a logic issue.

File: wm/client.go [hunk 25]
```diff
@@ -0,0 +1,236 @@
...
+type client struct {
+	id, win xproto.Window
+
+	iconic    bool
+	maximized bool
+
+	allowedActions []string
+	stateHints     []string
+
+	f  *frame
```
andydotxyz: This is accessed a lot from other code - it would be much clearer as "frame" - the duplication sucks but it is more readable.

wm/frame.go
```diff
@@ -3,238 +3,182 @@
...
 
-	f.wm.RaiseToTop(f)
+	f.c.wm.RaiseToTop(f.c)
 }
 
 func (f *frame) release(x, y int16) {
-	if x > int16(f.wm.borderWidth()) && x < int16(f.wm.borderWidth()+f.wm.buttonWidth()) &&
-		y < int16(f.wm.borderWidth()+f.wm.titleHeight()) {
-		f.Close()
+	ycoord := int16(f.c.wm.borderWidth() + f.c.wm.titleHeight())
```
andydotxyz: I think you missed my point here. Check at the top `if y > ycoord { return }`

wm/frame.go
andydotxyz: As for the 0, 0 - why create the variables at all if it's just 0, 0?

PR #20, opened 2019-09-02
Missing icon

okratitan:
This adds a missing icon when a .desktop match can not be found so that windows without matches still appear on the taskbar.  It also introduces hint handling for _NET_WM_SKIP_TASKBAR and sets this hint  on the root window children so that they skip the taskbar.

### DIFF
--- a/internal/fdo.go
+++ b/internal/fdo.go
@@ -119,7 +119,7 @@ func fdoLookupApplicationByMetadata(appName string) desktop.AppData {
 }
 
 //fdoLookupApplication looks up an application by name and returns an fdoApplicationData struct
-func fdoLookupApplication(appName string, fallbackIfMissing bool) desktop.AppData {
+func fdoLookupApplication(appName string) desktop.AppData {
 	locationLookup := fdoLookupXdgDataDirs()
 	for _, dataDir := range locationLookup {
 		testLocation := filepath.Join(dataDir, "applications", appName+".desktop")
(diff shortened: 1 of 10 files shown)


internal/fdo.go
```diff
@@ -118,7 +119,7 @@ func fdoLookupApplicationByMetadata(appName string) desktop.AppData {
 }
 
 //fdoLookupApplication looks up an application by name and returns an fdoApplicationData struct
-func fdoLookupApplication(appName string) desktop.AppData {
+func fdoLookupApplication(appName string, fallbackIfMissing bool) desktop.AppData {
```
andydotxyz: Every single call to this is `, false)` except once.
Therefore it would be neater to return it from the fdoLookupApplicationWinInfo func as a last resort.
Passing bool flags is usually a code smell...

internal/testdata/applications/app3.desktop
```diff
@@ -1,4 +1,4 @@
 [Desktop Entry]
 Name=App3
 Exec=app3
-Icon=app3
+Icon=/home/stephen/go/src/fyne.io/desktop/internal/testdata/icons/app3.png
```
andydotxyz: This looks like a mistaken commit - we need to fix this being overwritten all the time

okratitan: I don't think it breaks the mac files? Why would it? Also yes. Need to add that file to .gitignore

andydotxyz: Quite right, apologies I got confused between the two changes.
Strip this line and we're good

PR #24, opened 2019-10-01
Add the ability to move and resize windows not using Fyne borders.

### LINKED ISSUE #23: Support dragging chromium
Windows that do not have borders may implement their own drag, we should support this

okratitan:
Chromium, gtk3, and others use their own borders and titlebars and request movement and resizing via extended window manager hints.  This handles those hints to allow the window manager to move and resize those windows.

Fixes #23

### DIFF
--- a/wm/desk.go
+++ b/wm/desk.go
@@ -18,14 +18,33 @@ import (
 
 type x11WM struct {
 	stack
-	x      *xgbutil.XUtil
-	root   fyne.Window
-	rootID xproto.Window
-	loaded bool
+	x                   *xgbutil.XUtil
+	root                fyne.Window
+	rootID              xproto.Window
+	loaded              bool
+	moveResizing        bool
+	moveResizingStartX  int16
+	moveResizingStartY  int16
+	moveResizingGravity uint32
 
 	allowedActions []string
 	supportedHints []string
 }
+
+const (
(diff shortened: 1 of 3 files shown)


wm/desk.go
```diff
@@ -18,15 +18,34 @@ import (
...
-	x      *xgbutil.XUtil
-	root   fyne.Window
-	rootID xproto.Window
-	loaded bool
+	x                   *xgbutil.XUtil
+	root                fyne.Window
+	rootID              xproto.Window
+	loaded              bool
+	moveResizing        bool
+	moveResizingStartX  int16
```
andydotxyz: "Start" seems like a poor name as this is not atually the start value, more like the last or previous value.

wm/desk.go
```diff
@@ -18,15 +18,34 @@ import (
...
+	moveResizingStartY  int16
+	moveResizingGravity uint32
 
 	allowedActions []string
 	supportedHints []string
 }
 
+const (
```
andydotxyz: Could this be a named enum rater than uint32 - that indicates that only specific values rather than all uint32s would be allowed... (i.e. type moveResizeType uint32) 
Also is "gravity" a meaningful term here - if it can be a move instead of a resize is a more generic "type" more indicative?

PR #38, opened 2019-11-08
Clean up initial fullscreen and update border when title changes

okratitan:
This fixes the "flicker" that was seen when windows were initially set to fullscreen where the window would show up for a second and then go fullscreen.

This also now handles the property change notify hint so that we can start updating window properties as they change.  The first step is to update the window title when the name changes.

### DIFF
--- a/wm/frame.go
+++ b/wm/frame.go
@@ -359,8 +359,7 @@ func (f *frame) applyTheme(force bool) {
 	}
 }
 
-func (f *frame) setTitle() {
-	//Should we somehow just be changing the label rather than forcing a theme apply?
+func (f *frame) updateTitle() {
 	f.applyTheme(true)
 }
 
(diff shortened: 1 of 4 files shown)


wm/frame.go
```diff
@@ -356,6 +359,11 @@ func (f *frame) applyTheme(force bool) {
 	}
 }
 
+func (f *frame) setTitle() {
```
andydotxyz: A setter should take a parameter - this probably needs a different name?

wm/frame.go
```diff
@@ -356,6 +359,11 @@ func (f *frame) applyTheme(force bool) {
...
+	//Should we somehow just be changing the label rather than forcing a theme apply?
```
andydotxyz: This is not really a code document - more like a TODO or question, does it belong here?
If this is not the right way to do something should it be pushed?

p.s. Go likes a space after the slashes ;)

wm/frame.go
okratitan: Fixed.

wm/frame.go
okratitan: Removed the comment - After looking at the code is proper because applyTheme() handles refresh rather completely redoing the theme.

PR #39, opened 2019-11-11
Improve handling of borderless windows and hotswapping borders.

okratitan:
This can be seen via chrome/chromium.  In settings toggle Use system titlebar/borders.

### DIFF
--- a/wm/frame.go
+++ b/wm/frame.go
@@ -3,6 +3,7 @@
 package wm
 
 import (
+	"github.com/BurntSushi/xgbutil/ewmh"
 	"image"
 
 	"fyne.io/fyne"
@@ -22,7 +23,6 @@ type frame struct {
 	mouseX, mouseY          int16
 	resizeBottom            bool
 	resizeLeft, resizeRight bool
-	framed                  bool
 
 	minWidth, minHeight       uint
 	borderTop, borderTopRight xproto.Pixmap
@@ -32,10 +32,6 @@ type frame struct {
 }
 
 func (f *frame) unFrame() {
-	if !f.framed {
-		return
-	}
-
 	cli := f.client.wm.clientForWin(f.client.win)
 	if cli == nil {
 		return
@@ -97,7 +93,7 @@ func (f *frame) release(x, y int16) {
 	f.resizeBottom = false
 	f.resizeLeft = false
 	f.resizeRight = false
-	f.updateGeometry(f.x, f.y, f.width, f.height)
+	f.updateGeometry(f.x, f.y, f.width, f.height, false)
 }
 
 func (f *frame) drag(x, y int16) {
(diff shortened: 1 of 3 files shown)


wm/frame.go
```diff
@@ -342,21 +338,33 @@ func (f *frame) decorate(force bool) {
 		0, 0, int16(f.width-iconAndBorderSizePix), 0, uint16(iconAndBorderSizePix), uint16(iconSizePix))
 }
 
-func (f *frame) applyTheme(force bool) {
-	if !f.framed {
+func (f *frame) decorateBorderless() {
```
andydotxyz: This new method confuses me. Is it for borderless windows? if so isn't that the same as not decorated? so the chunk below is not executed?
After that it does a resize but only if it's fullscreen.
Can you help me understand? (possibly update the name to reflect what it does?)

wm/frame.go
```diff
@@ -388,6 +396,39 @@ func (f *frame) hide() {
 	xproto.UnmapWindow(f.client.wm.x.Conn(), f.client.win)
 }
 
+func (f *frame) changeBorder(){
```
andydotxyz: This name is a little generic - what does this method do? "applyBorder" or "setupBorder" seems likely candidates.
Also it's a very big method with 2 distinct functions - so probably it should contain just an if that then calls either "setupBorderFramed" and "setupBorderBorderless" or some such thing?
This then suggests an overlap with the decorateBorderless method above, but I didn't fully understand that so can't really say if this looks like duplication or not

wm/frame.go
```diff
@@ -184,29 +180,29 @@ func (f *frame) getGeometry() (int16, int16, uint16, uint16) {
 	return f.x, f.y, f.width, f.height
 }
 
-func (f *frame) updateGeometry(x, y int16, w, h uint16) {
-	resize := w != f.width || h != f.height
-	move := x != f.x || y != f.y
-	if !move && !resize {
-		return
+func (f *frame) updateGeometry(x, y int16, w, h uint16, force bool) {
```
andydotxyz: Passing bool here is indicating that the method does two things.
Perhaps a new method, "updateGeometryIfChanged" would do such a check, then call updateGeometry if there is a change?
This keeps the API a little clearer

wm/frame.go
okratitan: Yes it is for borderless windows.  I changed the name of the method to applyBorderlessTheme
This will applies the theme for borders/decorations if _MOTIF_WM_HINTS says so - or it will apply a completely borderless frame and theme (basically just a rectangle the same size as the window that is theme bg colored)

Fullscreen requires the additional resize, as was the case there in applyTheme() before this PR, but the additional resize is not necessary/will not render borderless correctly if not fullscreen and not decorated.

wm/frame.go
okratitan: I changed the name from changeBorder in client.c  to setupBorder.  setupBorder will then determine based on _MOTIF_WM_HINTS whether to addBorder() OR REMOVEbORDER()

wm/frame.go
okratitan: True - I followed the applyTheme method here of using force bool.  There are cases where we want to force a  geometry update (because otherwise it would determine the window has not moved or been resized and thus not apply geometry updates to the internal window).  What would you like to do with this?  Break it out into two methods as you suggest?  Or follow the pattern of applyTheme()?

PR #40, opened 2019-11-14
Implement click to focus.  Clicking anywhere on a window raises+focuses it now.

### LINKED ISSUE #32: Clicking a window in the background should raise it to top
This works on the window borders but not on window content

okratitan:
Fixes #32


andydotxyz:
Epic, thanks :)

PR #41, opened 2019-11-14
Use the window icon first when choosing border/taskbar icons.

### LINKED ISSUE #30: Window icons should use the window icon before falling back to the application icon

okratitan:
Fixes #30

### DIFF
--- a/wm/client.go
+++ b/wm/client.go
@@ -66,12 +66,11 @@ func (c *client) IconName() string {
 }
 
 func (c *client) Icon() fyne.Resource {
-	var icon fyne.Resource = nil
 	xIcon := windowIcon(c.wm.x, c.win, iconSize, iconSize)
-	if xIcon.Bytes() != nil {
-		icon = fyne.NewStaticResource(c.Title(), xIcon.Bytes())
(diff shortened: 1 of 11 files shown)


andydotxyz:
Works well and code is clean. Happy to accept, however as discussed on Slack should we perhaps fall back to theme or app in the taskbar so it's not all pixelated?

andydotxyz:
oh man I think there is something awry with the nil checking - we now get missing icons instead of the missing icon icon...

wm/client.go
```diff
@@ -65,6 +65,15 @@ func (c *client) IconName() string {
 	return windowIconName(c.wm.x, c.win)
 }
 
+func (c *client) Icon() fyne.Resource {
+	var icon fyne.Resource = nil
```
andydotxyz: This might be your problem... try moving the return into the if statement - the icon var is not needed...

okratitan:
Of note - I think transparency may be broken... It's weird.

andydotxyz:
Yup, good to merge as soon as Travis is happy (gofmt error)

PR #44, opened 2019-12-06
Implement XRandR Multihead Support - Fix Gnome Apps - Scale Desktops

### LINKED ISSUE #27: Add multiple monitor support

okratitan:
This supports configurable primary screens, user defined scaling, auto scaling, and fixes a lot of issues with apps such as gnome apps.

The one caveat here is that there is one scale value across all monitors, rather than being calculated per monitor.  This will need to be addressed in the future.

Fixes #27

### DIFF
--- a/cmd/fynedesk/main.go
+++ b/cmd/fynedesk/main.go
@@ -1,27 +1,12 @@
 package main
 
 import (
-	"os"
-
-	"fyne.io/fyne"
 	"fyne.io/fyne/app"
 )
-
-func startSettingsListener(app fyne.App, listener chan fyne.Settings) {
-	for {
-		_ = <-listener
-		if os.Getenv("FYNE_DESK_RUNNER") != "" {
-			os.Exit(1)
-		}
-	}
-}
 
 func main() {
 	a := app.NewWithID("io.fyne.desktop")
 	desk := setupDesktop(a)
-	listener := make(chan fyne.Settings)
-	a.Settings().AddChangeListener(listener)
-	go startSettingsListener(a, listener)
 
 	desk.Run()
 }
--- a/cmd/fynedesk/main.go
+++ b/cmd/fynedesk/main.go
@@ -1,12 +1,27 @@
 package main
 
 import (
+	"os"
+
+	"fyne.io/fyne"
 	"fyne.io/fyne/app"
 )
+
+func startSettingsListener(app fyne.App, listener chan fyne.Settings) {
+	for {
+		_ = <-listener
+		if os.Getenv("FYNE_DESK_RUNNER") != "" {
+			os.Exit(1)
+		}
+	}
+}
 
 func main() {
 	a := app.NewWithID("io.fyne.desktop")
 	desk := setupDesktop(a)
+	listener := make(chan fyne.Settings)
+	a.Settings().AddChangeListener(listener)
+	go startSettingsListener(a, listener)
 
 	desk.Run()
 }
--- a/cmd/fynedesk/main_linux.go
+++ b/cmd/fynedesk/main_linux.go
@@ -14,12 +14,12 @@ import (
 
 func setupDesktop(a fyne.App) desktop.Desktop {
 	icons := internal.NewFDOIconProvider()
-	mgr, err := wm.NewX11WindowManager(a)
+	mgr, screensProvider, err := wm.NewX11WindowManager(a)
 	if err != nil {
 		log.Println("Could not create window manager:", err)
 		return desktop.NewEmbeddedDesktop(a, icons)
 	}
-	desk := desktop.NewDesktop(a, mgr, icons, mgr.(desktop.Screens))
+	desk := desktop.NewDesktop(a, mgr, icons, screensProvider)
 	mgr.SetRoot(desk.Root())
 	return desk
 }
--- a/cmd/fynedesk/main_linux.go
+++ b/cmd/fynedesk/main_linux.go
@@ -14,12 +14,12 @@ import (
 
 func setupDesktop(a fyne.App) desktop.Desktop {
 	icons := internal.NewFDOIconProvider()
-	mgr, screensProvider, err := wm.NewX11WindowManager(a)
+	mgr, err := wm.NewX11WindowManager(a)
 	if err != nil {
 		log.Println("Could not create window manager:", err)
 		return desktop.NewEmbeddedDesktop(a, icons)
 	}
-	desk := desktop.NewDesktop(a, mgr, icons, screensProvider)
+	desk := desktop.NewDesktop(a, mgr, icons, wm.NewX11ScreensProvider(mgr))
 	mgr.SetRoot(desk.Root())
 	return desk
 }
--- a/cmd/fynedesk/main_linux.go
+++ b/cmd/fynedesk/main_linux.go
@@ -19,8 +19,7 @@ func setupDesktop(a fyne.App) desktop.Desktop {
 		log.Println("Could not create window manager:", err)
 		return desktop.NewEmbeddedDesktop(a, icons)
 	}
-
-	desk := desktop.NewDesktop(a, mgr, icons)
+	desk := desktop.NewDesktop(a, mgr, icons, mgr.(desktop.Screens))
 	mgr.SetRoot(desk.Root())
 	return desk
 }
--- a/desk.go
+++ b/desk.go
@@ -9,13 +9,13 @@ import (
 	"fyne.io/fyne"
 )
 
-// Desktop defines an embedded or full desktop envionment that we can run.
+// Desktop defines an embedded or full desktop environment that we can run.
 type Desktop interface {
 	Root() fyne.Window
 	Run()
 	RunApp(AppData) error
 	Settings() DeskSettings
-	ContentSizePixels() (uint32, uint32)
+	ContentSizePixels(screenIndex int) (uint32, uint32)
 
 	IconProvider() ApplicationProvider
 	WindowManager() WindowManager
@@ -28,21 +28,46 @@ type deskLayout struct {
 	win      fyne.Window
 	wm       WindowManager
 	icons    ApplicationProvider
+	screens  Screens
 	settings DeskSettings
 
-	background, bar, widgets, mouse fyne.CanvasObject
+	backgrounds         []fyne.CanvasObject
+	bar, widgets, mouse fyne.CanvasObject
+	container           *fyne.Container
 }
 
 func (l *deskLayout) Layout(objs []fyne.CanvasObject, size fyne.Size) {
+	var x, y, w, h, index int = 0, 0, 0, 0, 0
+
+	if l.screens.Screens() != nil && len(l.screens.Screens()) > 1 && len(l.backgrounds) > 1 {
+		screens := l.screens.Screens()
+		index = l.screens.Primary()
+		x, y, w, h = screens[index].ScaledX, screens[index].ScaledY,
+			screens[index].ScaledWidth, screens[index].ScaledHeight
+		size.Width = w
+		size.Height = h
+
+		for i := 0; i < len(screens); i++ {
+			if i == index {
+				continue
+			}
+			xx, yy, ww, hh := screens[i].ScaledX, screens[i].ScaledY, screens[i].ScaledWidth, screens[i].ScaledHeight
+
+			l.backgrounds[i].Move(fyne.NewPos(xx, yy))
+			l.backgrounds[i].Resize(fyne.NewSize(ww, hh))
+		}
+	}
+
 	barHeight := l.bar.MinSize().Height
-	l.bar.Resize(fyne.NewSize(size.Width, barHeight))
-	l.bar.Move(fyne.NewPos(0, size.Height-barHeight))
+	l.bar.Resize(fyne.NewSize(size.Width, y+barHeight))
+	l.bar.Move(fyne.NewPos(x, y+size.Height-barHeight))
 
 	widgetsWidth := l.widgets.MinSize().Width
 	l.widgets.Resize(fyne.NewSize(widgetsWidth, size.Height))
-	l.widgets.Move(fyne.NewPos(size.Width-widgetsWidth, 0))
+	l.widgets.Move(fyne.NewPos(x+size.Width-widgetsWidth, y))
 
-	l.background.Resize(size)
+	l.backgrounds[index].Move(fyne.NewPos(x, y))
+	l.backgrounds[index].Resize(size)
 }
 
 func (l *deskLayout) MinSize(objects []fyne.CanvasObject) fyne.Size {
@@ -64,17 +89,22 @@ func (l *deskLayout) Root() fyne.Window {
 	if l.win == nil {
 		l.win = l.newDesktopWindow()
 
-		l.background = newBackground()
+		l.backgrounds = append(l.backgrounds, newBackground())
 		l.bar = newBar(l)
 		l.widgets = newWidgetPanel(l)
 		l.mouse = newMouse()
-		l.win.SetContent(fyne.NewContainerWithLayout(l,
-			l.background,
-			l.bar,
-			l.widgets,
-			mouse,
-		))
+		l.container = fyne.NewContainerWithLayout(l, l.backgrounds[0])
+		if l.screens.Screens() != nil && len(l.screens.Screens()) > 1 {
+			for i := 1; i < len(l.screens.Screens()); i++ {
+				l.backgrounds = append(l.backgrounds, newBackground())
+				l.container.AddObject(l.backgrounds[i])
+			}
+		}
+		l.container.AddObject(l.bar)
+		l.container.AddObject(l.widgets)
+		l.container.AddObject(mouse)
 
+		l.win.SetContent(l.container)
 		l.mouse.Hide() // temporarily we do not handle mouse (using X default)
 		if l.wm != nil {
 			l.win.SetOnClosed(func() {
@@ -110,10 +140,16 @@ func (l *deskLayout) Settings() DeskSettings {
 	return l.settings
 }
 
-func (l *deskLayout) ContentSizePixels() (uint32, uint32) {
-	screenW := uint32(float32(l.background.Size().Width) * l.Root().Canvas().Scale())
-	screenH := uint32(float32(l.background.Size().Height) * l.Root().Canvas().Scale())
-	return screenW - uint32(float32(l.widgets.Size().Width)*l.Root().Canvas().Scale()), screenH
+func (l *deskLayout) ContentSizePixels(screenIndex int) (uint32, uint32) {
+	if l.backgrounds[screenIndex] != nil {
+		screenW := uint32(l.backgrounds[screenIndex].Size().Width)
+		screenH := uint32(l.backgrounds[screenIndex].Size().Height)
+		if l.screens.Primary() == screenIndex {
+			return screenW - uint32(float32(l.widgets.Size().Width)*l.Root().Canvas().Scale()), screenH
+		}
+		return screenW, screenH
+	}
+	return 0, 0
 }
 
 func (l *deskLayout) IconProvider() ApplicationProvider {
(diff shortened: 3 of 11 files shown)


andydotxyz:
Unfortunately maximize is not working when scale != 1.
Travis is very unhappy https://travis-ci.org/fyne-io/desktop - it seems that platform independent code now depends on X11 stuff that is not compiled in for testing - run the tests locally with "ci" flag set to replicate it.

wm/frame.go
```diff
@@ -221,12 +221,26 @@ func (f *frame) maximizeApply() {
 	f.client.restoreX = f.x
 	f.client.restoreY = f.y
 
-	maxWidth, maxHeight := desktop.Instance().ContentSizePixels()
+	var screenIndex, startX, startY int = 0, 0, 0
+	screens := f.client.wm.Screens()
+	if len(screens) > 1 {
```
andydotxyz: Shouldn't this be in a helpful function (like ScreenForWindow) somewhere like the Screens interface?

wm/frame.go
```diff
@@ -451,10 +462,10 @@ func newFrame(c *client) *frame {
 	full := c.Fullscreened()
 	decorated := c.Decorated()
 	if full {
-		x = 0
-		y = 0
-		w = c.wm.x.Screen().WidthInPixels
-		h = c.wm.x.Screen().HeightInPixels
+		x = int16(c.wm.Screens()[c.wm.Active()].X)
```
andydotxyz: This is a lot of lookups to set 4 variables, perhaps the API changes above make it cleaner?

wm/screens.go
```diff
@@ -0,0 +1,123 @@
...
+				screen.ScaledHeight = int(float32(screen.Height) / scale)
+				x.screens[j] = screen
+			}
+		}
+	}
+	if len(x.screens) == 0 {
+		x.screens = append(x.screens, desktop.Screen{Name: "Screen0", Index: 0,
+			X: xwindow.RootGeometry(x.x).X(), Y: xwindow.RootGeometry(x.x).Y(),
+			Width: xwindow.RootGeometry(x.x).Width(), Height: xwindow.RootGeometry(x.x).Height(),
+			ScaledX: xwindow.RootGeometry(x.x).X(), ScaledY: xwindow.RootGeometry(x.x).Y(),
```
andydotxyz: The values being set here into the scaled fields are not scaled?

wm/screens.go
```diff
@@ -0,0 +1,123 @@
...
+			ScaledWidth: xwindow.RootGeometry(x.x).Width(), ScaledHeight: xwindow.RootGeometry(x.x).Height()})
+		x.primary = 0
```
andydotxyz: zero values don't need to be set do they?

wm/stack.go
```diff
@@ -1,19 +1,25 @@
...
-import "fyne.io/desktop"
+import (
+	"fyne.io/desktop"
+	"github.com/BurntSushi/xgb/xproto"
+)
 
 type stack struct {
-	clients []desktop.Window
+	clients      []desktop.Window
+	mappingOrder []desktop.Window
```
andydotxyz: I don't understand what mappingOrder is here - it doesn't seem like it's part of stack functionality which is reinforced by it requiring the addition of xgb dependency to a high level concept.

wm/stack.go
```diff
@@ -28,6 +34,29 @@ func (s *stack) removeFromStack(win desktop.Window) {
...
+		return
+	}
+	s.mappingOrder = append(s.mappingOrder[:pos], s.mappingOrder[pos+1:]...)
+}
+
+func (s *stack) getMappingOrder() []xproto.Window {
+	return s.getWindowsFromClients(s.mappingOrder)
+}
+
+func (s *stack) getWindowsFromClients(clients []desktop.Window) []xproto.Window {
```
andydotxyz: This doesn't seem to belong as a function on stack - it does not use any of it's state... perhaps it's another helper function within wm/desk or wm/client?

desk.go
```diff
@@ -110,10 +140,16 @@ func (l *deskLayout) Settings() DeskSettings {
 	return l.settings
 }
 
-func (l *deskLayout) ContentSizePixels() (uint32, uint32) {
-	screenW := uint32(float32(l.background.Size().Width) * l.Root().Canvas().Scale())
-	screenH := uint32(float32(l.background.Size().Height) * l.Root().Canvas().Scale())
-	return screenW - uint32(float32(l.widgets.Size().Width)*l.Root().Canvas().Scale()), screenH
+func (l *deskLayout) ContentSizePixels(screenIndex int) (uint32, uint32) {
```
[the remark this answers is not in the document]
okratitan: Fixed

desk.go
```diff
@@ -9,13 +9,13 @@ import (
...
 
-// Desktop defines an embedded or full desktop envionment that we can run.
+// Desktop defines an embedded or full desktop environment that we can run.
 type Desktop interface {
 	Root() fyne.Window
 	Run()
 	RunApp(AppData) error
 	Settings() DeskSettings
-	ContentSizePixels() (uint32, uint32)
+	ContentSizePixels(screenIndex int) (uint32, uint32)
```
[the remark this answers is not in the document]
okratitan: Fixed to pass the Screen type

screen.go
```diff
@@ -0,0 +1,17 @@
...
+type Screens interface {
+	Screens() []Screen // Screens returns a Screen type slice of each available physical screen
+	Active() int       // Active returns the screen index of the currently active screen
+	Primary() int      // Primary returns the screen index of the primary screen
+}
+
+// Screen provides relative information about a single physical screen
+type Screen struct {
+	Name                      string // Name is the randr provided name of the screen
+	Index                     int    // Index is the position of the screen in the screens slice.
```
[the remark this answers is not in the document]
okratitan: Agreed. Removed.

screen.go
```diff
@@ -0,0 +1,17 @@
+package desktop
+
+// Screens provides information about available physical screens for Fyne desktop
+type Screens interface {
+	Screens() []Screen // Screens returns a Screen type slice of each available physical screen
```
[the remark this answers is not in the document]
okratitan: Agreed.  I moved the "Screen" type to "Head" so that we have Screens and Head

screen.go
```diff
@@ -0,0 +1,17 @@
...
+	Active() int       // Active returns the screen index of the currently active screen
```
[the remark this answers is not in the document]
okratitan: Yes - Fixed.

screen.go
```diff
@@ -0,0 +1,17 @@
...
+	Primary() int      // Primary returns the screen index of the primary screen
+}
+
+// Screen provides relative information about a single physical screen
+type Screen struct {
+	Name                      string // Name is the randr provided name of the screen
+	Index                     int    // Index is the position of the screen in the screens slice.
+	X, Y, Width, Height       int    // Geometry of the screen
+	ScaledX, ScaledY          int    // Scaled position of the screen
```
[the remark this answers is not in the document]
okratitan: So - If the scaling is done based off the screen's information - Layout will not have access to that as it requires XRandR specifics.

wm/desk.go
```diff
@@ -552,6 +583,8 @@ func (x *x11WM) setupWindow(win xproto.Window) {
 
 	x.AddWindow(c)
 	c.Focus()
+	windowClientListSet(x.x, x.getMappingOrder())
```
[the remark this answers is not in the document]
okratitan: Yes - Changed to updates.

wm/property.go
```diff
@@ -158,3 +158,11 @@ func windowExtendedHintsRemove(x *xgbutil.XUtil, win xproto.Window, hint string)
 		}
 	}
 }
+
+func windowClientListSet(x *xgbutil.XUtil, wins []xproto.Window) {
```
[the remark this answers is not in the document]
okratitan: Because the functions are in the property.go file as they are ewmh settings and it follows the convention of every single function in propoerty.go starting with window.  We can revisit that?

desk.go
```diff
@@ -134,6 +170,24 @@ func (l *deskLayout) scaleVars(scale float32) []string {
 	}
 }
 
+func (l *deskLayout) Screens() []Screen {
```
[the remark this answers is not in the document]
okratitan: Because in the case of an embedded desktop - the deskLayout won't have access to x11wm information

So depending on if it is embedded desktop vs regular - we set which interface to use - which is cleaner in the rest of the code to say deskLayout.screens rather than if (embedded) everywhere.

cmd/fynedesk/main_linux.go
```diff
@@ -19,8 +19,7 @@ func setupDesktop(a fyne.App) desktop.Desktop {
 		log.Println("Could not create window manager:", err)
 		return desktop.NewEmbeddedDesktop(a, icons)
 	}
-
-	desk := desktop.NewDesktop(a, mgr, icons)
+	desk := desktop.NewDesktop(a, mgr, icons, mgr.(desktop.Screens))
```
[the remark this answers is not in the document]
okratitan: Because mgr here is WindowManager and the interface is on x11WM

desk.go
```diff
@@ -153,5 +207,6 @@ func NewDesktop(app fyne.App, wm WindowManager, icons ApplicationProvider) Deskt
 // fyne/test package.
 func NewEmbeddedDesktop(app fyne.App, icons ApplicationProvider) Desktop {
 	instance = &deskLayout{app: app, icons: icons, settings: NewDeskSettings()}
+	instance.(*deskLayout).screens = instance.(*deskLayout)
```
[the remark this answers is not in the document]
okratitan: Because in the case of an embedded desktop - the deskLayout won't have access to x11wm information

So depending on if it is embedded desktop vs regular - we set which interface to use - which is cleaner in the rest of the code to say deskLayout.screens rather than if (embedded) everywhere.

desk_test.go
```diff
@@ -53,14 +53,15 @@ func (*testSettings) Background() string {
 
 func TestDeskLayout_Layout(t *testing.T) {
 	l := &deskLayout{}
-	l.background = canvas.NewRectangle(color.White)
+	l.screens = l
```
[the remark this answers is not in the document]
okratitan: Both - cleaner in my opinion due to reasons explained above.

wm/client.go
```diff
@@ -318,22 +319,14 @@ func newClient(win xproto.Window, wm *x11WM) *client {
 	}
 	windowAllowedActionsSet(wm.x, win, wm.allowedActions)
 	initialHints := windowExtendedHintsGet(c.wm.x, c.win)
-	removeHints := wm.pendingRemoveHints[win]
```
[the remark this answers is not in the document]
okratitan: Yes - removeHints waited on the client to be created to remove the hint - but in the case of things like _NET_WM_HIDDEN - this wasn'r proper as a client was never going to be created without showing the window.  And showing the window creates the client anyway.

wm/desk.go
```diff
@@ -431,8 +450,8 @@ func (x *x11WM) handleStateActionRequest(ev xproto.ClientMessageEvent, removeSta
 func (x *x11WM) handleInitialHints(ev xproto.ClientMessageEvent, hint string) {
 	switch clientMessageStateAction(ev.Data.Data32[0]) {
 	case clientMessageStateActionRemove:
-		hints := x.pendingRemoveHints[ev.Window]
-		hints = append(hints, hint)
+		windowExtendedHintsRemove(x.x, ev.Window, hint)
+		x.showWindow(ev.Window)
```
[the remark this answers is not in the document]
okratitan: yes - showWindow is required because the window is hidden and _NET_WM_HIDDEN remove is telling us to show the window... so showWindow will setup a client by calling setupWindow and setupWindow calls for a new frame which shows the window.

wm/desk.go
```diff
@@ -518,6 +537,18 @@ func (x *x11WM) showWindow(win xproto.Window) {
 	if x.rootID == 0 {
 		return
 	}
+	hints, err := icccm.WmHintsGet(x.x, win)
+	if err != nil {
+		fyne.LogError("Could not get initial hints for client", err)
+	} else {
```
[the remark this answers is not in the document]
okratitan: Indeed yes

wm/frame.go
okratitan: Indeed. Added.

wm/screens.go
okratitan: Correct - If we found no screen information from xrandr, then we have nothing to calculate scale based off of here.

wm/stack.go
okratitan: Extended window manager hints requires two lists... a list of clients based on the order they were added to the stack and a list of clients based on the current position in the stack.  I have removed the xgb dependency.

wm/stack.go
okratitan: Moved to Client.

wm/frame.go
okratitan: indeed. made it one lookup.

wm/screens.go
okratitan: Removed since handling screens off of indexes is no longer applicable.

desk_test.go
```diff
@@ -53,14 +53,15 @@ func (*testSettings) Background() string {
...
+	l.backgrounds = append(l.backgrounds, canvas.NewRectangle(color.White))
```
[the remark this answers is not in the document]
okratitan: l.backgrounds expects a slice of []fyne.CanvasObject - perhaps because I'm newish to Go I'm not sure how to assign an array of [1]fyne.CanvasObject to that?  I get errors.

cmd/fynedesk/main.go
```diff
@@ -1,12 +1,27 @@
...
+	"os"
+
+	"fyne.io/fyne"
 	"fyne.io/fyne/app"
 )
 
+func startSettingsListener(app fyne.App, listener chan fyne.Settings) {
+	for {
+		_ = <-listener
+		if os.Getenv("FYNE_DESK_RUNNER") != "" {
```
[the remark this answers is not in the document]
okratitan: The issue here - and we can revisit this over the next few days - is how to do the refresh when it will require access to multiple packages that don't have access to each other.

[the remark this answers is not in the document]
andydotxyz: This listener could be put in a non-main package

desk_test.go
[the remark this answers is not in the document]
andydotxyz: Yes I see that the layout contains a Screens type,  but I don’t understand why there is the additional implementation of the interface in the layout code here...

screen.go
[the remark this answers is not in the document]
andydotxyz: I don’t know if “ahead” will mean much to people. Also it seemed that “Screens()” was the duplication rather than the “Screen” type

desk_test.go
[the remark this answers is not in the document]
andydotxyz: Sorry, I don’t understand your argument for implementing in the containing type

desk.go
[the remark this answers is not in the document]
andydotxyz: Yes I see that the layout contains a Screens type, but I don’t understand why there is the additional implementation of the interface in the layout code here...

desk_test.go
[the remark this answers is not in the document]
andydotxyz: []fyne.CanvasObject{element1}

wm/stack.go
andydotxyz: If this truly is a requirement of the stack functionality then it needs appropriate tests. From what you are describing though it seems like an x11 specific requirement for storing mapping order which is quite separate from stacking order.

andydotxyz:
The latest version (Probably due to some alterations to full screen checks) introduces a new issue, seen with any app that tries to remember maximised state.

Using goland:
* open main window
* maximise 
* iconofy
* uniconify

Now notice that it cannot be dragged.

* restart goland

Notice it still cannot be dragged

* maximise

You will see it’s 0x0 in the top left

screen.go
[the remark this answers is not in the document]
okratitan: Sure I can go with ScreenList and Screen

wm/stack.go
okratitan: I would imagine other wm specs might want this too... but the simple answer is... stack manages our list of clients - when clients are added when they are removed - etc... it felt like the wrong place to add a client list anywhere else as that was already the functionality of stack.

desk.go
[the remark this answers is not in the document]
okratitan: Because of embedded desktop ... the screens interface from x11WM will not work with embedded desktop mode, so something else has to implement the interface as well.  In this case that something is deskLayout.

wm/desk.go
```diff
@@ -32,17 +32,24 @@ type x11WM struct {
 	moveResizingType  moveResizeType
 	altTabList        []desktop.Window
 	altTabIndex       int
-	screens           []*desktop.Head
-	active            *desktop.Head
-	primary           *desktop.Head
+	screens           []*desktop.Screen
```
andydotxyz: Don't you think that the state for a screens provider should be in the screens provider?
The instance below is only needed because you haven't encapsulated the data appropriately.

wm/desk.go
```diff
@@ -32,17 +32,24 @@ type x11WM struct {
...
+	active            *desktop.Screen
+	primary           *desktop.Screen
 	scale             float32
 
 	allowedActions []string
 	supportedHints []string
+
+	screensProvider desktop.ScreenList
```
andydotxyz: This should not be needed - the screenlist is handled by the desktop code in the root package - this seems to only be set in the constructor and then passed out again - so no state in the x11WM type needed...

cmd/fynedesk/main_linux.go
```diff
@@ -14,12 +14,12 @@ import (
 
 func setupDesktop(a fyne.App) desktop.Desktop {
 	icons := internal.NewFDOIconProvider()
-	mgr, err := wm.NewX11WindowManager(a)
+	mgr, screensProvider, err := wm.NewX11WindowManager(a)
```
andydotxyz: Given that you have created a new constructor, NewX11ScreensProvider, why does this have to be returned by a different constructor?

wm/desk.go
```diff
@@ -83,44 +90,52 @@ func (x *x11WM) Blank() {
...
+}
+
+// NewX11ScreensProvider returns a screen provider for use in x11 desktop mode
+func NewX11ScreensProvider() desktop.ScreenList {
+	return &x11ScreensProvider{}
 }
 
 // NewX11WindowManager sets up a new X11 Window Manager to control a desktop in X11.
-func NewX11WindowManager(a fyne.App) (desktop.WindowManager, error) {
+func NewX11WindowManager(a fyne.App) (desktop.WindowManager, desktop.ScreenList, error) {
```
andydotxyz: returning multiple types from a constructor func is confusing - and as noted above this should not be needed as the x11ScreensProvider has it's own constructor

wm/frame.go
```diff
@@ -248,6 +248,11 @@ func (f *frame) maximizeApply() {
 }
 
 func (f *frame) unmaximizeApply() {
+	if f.client.restoreWidth == 0 && f.client.restoreHeight == 0 {
+		screen := f.client.wm.screensProvider.ScreenForWindow(int(f.client.restoreX), int(f.client.restoreY))
```
andydotxyz: What's the point in passing in the restore parameters - if the width and height are 0 then this is going to be a pretty redundant call... surely the active screen would be more appropriate?

screen.go
```diff
@@ -1,16 +1,16 @@
...
-	Primary() *Head                                 // Primary returns the screen index of the primary screen
-	Scale() float32                                 // Return the scale calculated for use across screens
-	ScreenForWindow(windowX int, windowY int) *Head // Return the screen that a window is located on
+// ScreenList provides information about available physical screens for Fyne desktop
+type ScreenList interface {
+	Screens() []*Screen                               // Screens returns a Screen type slice of each available physical screen
+	Active() *Screen                                  // Active returns the screen index of the currently active screen
+	Primary() *Screen                                 // Primary returns the screen index of the primary screen
+	Scale() float32                                   // Return the scale calculated for use across screens
+	ScreenForWindow(windowX int, windowY int) *Screen // Return the screen that a window is located on
```
andydotxyz: Given that this is a typed interface it should probably be passing a *Window - then the list logic can determine which part of the window is an anchor point for determining which screen...
(as discussed on slack this would probably be better as the centre of the window)

wm/screens.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+	}
+	dpi := float32(widthPx) / (float32(widthMm) / 25.4)
+	if dpi > 1000 || dpi < 10 {
+		dpi = 96
+	}
+	return float32(math.Round(float64(dpi)/144.0*10.0)) / 10.0
+}
+
+func (x *x11WM) setupScreens() {
```
andydotxyz: adding !ci to the build tags on this file would fix travis.
Also why isn't the screens provider definition in this file, with the following function (appropriately tidied for readability) on that type instead of the x11WM type?

wm/frame.go
okratitan: Being safe and covering a case where x, y may exist (i.e. top left or corner of a second screen) but w, h are 0

screen.go
okratitan: Agreed.

wm/desk.go
okratitan: Fixed

cmd/fynedesk/main_linux.go
okratitan: Fixed.

wm/desk.go
okratitan: Fixed.

wm/screens.go
okratitan: Done.

wm/desk.go
okratitan: yes

PR #45, opened 2019-12-11
Adjust windows according to their desired increment.  Notify windows when their size changes

okratitan:
This follows the icccm spec and also fixes applications like terminology.

### DIFF
--- a/wm/frame.go
+++ b/wm/frame.go
@@ -3,8 +3,9 @@
 package wm
 
 import (
-	"github.com/BurntSushi/xgbutil/ewmh"
 	"image"
+
+	"github.com/BurntSushi/xgbutil/ewmh"
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/theme"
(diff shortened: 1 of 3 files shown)


wm/property.go
```diff
@@ -103,6 +104,34 @@ func windowMinSize(x *xgbutil.XUtil, win xproto.Window) (uint, uint) {
 	return 0, 0
 }
 
+func windowApplyIncrement(x *xgbutil.XUtil, win xproto.Window, width uint16, height uint16) (uint16, uint16) {
```
andydotxyz: This is not actually applying anything - it is returning a value so verbs like calculate, round or similar would be clearer.
Perhaps even windowSizeWithIncrement?

okratitan: updated to be windowSizeWithIncrement

okratitan:
Okay the second commit further improves our window handling.  It follows requested resize increments, and also alerts clients to their geometry according to the icccm spec so that they now know where they are located and where to place their sub menus/windows.  This also fixes the reframing on fyndesk_runner reload to where the windows are no longer offset but kept in their same position.

You will now see gnome-terminal, lxterminal, terminology, goland, etc..., etc... all working the way they should.

andydotxyz:
Very close. The reparent on restart is a great fix.
However when you drag a window after a restart it then jumps to the offset position it would have been at before.

wm/frame.go
```diff
@@ -397,7 +411,10 @@ func (f *frame) updateTitle() {
 }
 
 func (f *frame) borderWidth() uint16 {
-	return f.client.wm.scaleToPixels(wmTheme.BorderWidth)
+	if f.client.Decorated() {
```
andydotxyz: This logic seems the wrong way round. The main content of a function should be at the top level.
So if !decorated { return 0 } would read more logically

PR #46, opened 2019-12-12
Fix dialogs hanging windows and fix iconified windows on restart.

okratitan:
Dialogs that were transient - yet requested decoration were hanging windows.  This fixes that.  Further this handles iconified windows if the window is requested to start iconified or if fynedesk_runner reloads the desktop.


wm/desk.go
```diff
@@ -526,15 +526,15 @@ func (x *x11WM) showWindow(win xproto.Window) {
...
-	} else if (hints.Flags & xproto.CwOverrideRedirect) != 0 {
-		return
-	}
-	transient, err := icccm.WmTransientForGet(x.x, win)
+	transient := windowTransientForGet(x.x, win)
 	if transient != 0 {
-		return
+		winType := windowTypeGet(x.x, win)
+		switch winType[0] {
+		case "_NET_WM_WINDOW_TYPE_UTILITY", "_NET_WM_WINDOW_TYPE_DIALOG", "_NET_WM_WINDOW_TYPE_NORMAL":
```
andydotxyz: These should probably be defined as consts - especially as it is used again  in wm/property.go

Issue #15, opened 2019-08-20
Add ability to configure app bar


okratitan:
Ability to turn zooming on or off
Set default icons (ala a search that works similar to app exec in panel?)
Set default icon theme
Set default icon size
Turn taskbar functionality on or off maybe?
Change zoom scale maybe?

PR #51, opened 2019-12-16
Add configuration for background, icon theme, and default icons

### LINKED ISSUE #14: Add ability to configure desktop background
Loch fyne is awesome! So are other bgs!

okratitan:
Fixes #14 and #15


andydotxyz:
This has fixes from other PRs in it as well. Can you make a clean one based on latest develop to clean up the various other things in it's history?

andydotxyz:
Also there is a lot of parsing of string lists with "|" separator in here.
Why not simply store in the API as a []string so that it is not parsed in and out all the time.
The setter can then simply set it to strings.Join(defaultApps, "|") and the only parsing you have to do is on loading the config.

okratitan:
Alright cleaned and change it to store as []string

okratitan:
Yes currently these settings require you to reload fynedesk... Background changing would be the easiest one to add a way to live update.

Adding/Removing default apps will be a little more difficult - but not bad - you will just need to keep a list of the default apps in the app bar, likely in a map to their app name, so you can remove that object then Refresh() or in the case of adding an icon you will need to know the position of the separator and add it before the separator, or in the case of us handling order (which we should) you will have to insert it in the correct spot.

Icon Theme - I'm not sure how to do this at the moment - if it was just the default app icons that needed to be updated, it would be simple - but we would also need to be able to update the icons in the taskbar, the icons in window border, and we can't just clear the bar and readd the default apps or we would lose the taskbar apps in the process.  I will have to give that thought.

okratitan:
Also the UI needs to be better than what I have here.  Lets discuss what it should be - As mentioned in Slack - if you have more icon themes than the settings window size can handle - select widget does not scroll so it doesn't work.

On the default app icon side of it ... we really should allow order setting and have more than just a blank screen waiting for text in the entry.  We need to be able to handle both lookup and order setting.  We also could just list every single app ... I do have the api there now... IconProvider().AllApps() but the scrolling as it was - was really bad.  If we listed all apps - we could make sure the default apps were all at the top in correct order and allow changing of those via Tilo's method? The default apps would be the ones with checks... and then all the unchecked apps would be below that in the list.

okratitan:
Anyhow - Just thought I would leave my thoughts/opinions here and get feedback.

bar.go
```diff
@@ -89,7 +89,8 @@ func newBar(desk Desktop) fyne.CanvasObject {
 	if desk.WindowManager() != nil {
 		desk.WindowManager().AddStackListener(appBar)
 	}
-	for _, app := range appBar.desk.IconProvider().DefaultApps() {
+	for _, name := range Instance().Settings().DefaultApps() {
```
andydotxyz: I realise in the morning light that there are too many Instance() lookups - we should try and avoid them.
Right above us we are passed the desktop instance as a variable so that should be used.

bar.go
```diff
@@ -89,7 +89,8 @@ func newBar(desk Desktop) fyne.CanvasObject {
...
+		app := appBar.desk.IconProvider().FindAppFromName(name)
```
andydotxyz: For safety we should check that the app was found - it's possible that someone could edit the file (or sync it ;) ) and end up with a nil app here...

desk.go
```diff
@@ -246,7 +246,8 @@ func Instance() Desktop {
 // The WindowManager passed in will be used to manage the screen it is loaded on.
 // An ApplicationProvider is used to lookup application icons from the operating system.
 func NewDesktop(app fyne.App, wm WindowManager, icons ApplicationProvider, screenProvider ScreenList) Desktop {
-	instance = &deskLayout{app: app, wm: wm, icons: icons, screens: screenProvider, settings: NewDeskSettings()}
+	instance = &deskLayout{app: app, wm: wm, icons: icons, screens: screenProvider}
```
andydotxyz: I'm not sure what this file has changed for, can you help?

icons.go
```diff
@@ -12,8 +12,10 @@ type AppData interface {
 
 // ApplicationProvider describes a type that can locate icons and applications for the current system
 type ApplicationProvider interface {
+	AllApps() []AppData
+	AvailableThemes() []string
 	FindAppFromName(appName string) AppData
 	FindAppFromWinInfo(win Window) AppData
 	FindAppsMatching(pattern string) []AppData
-	DefaultApps() []AppData
```
andydotxyz: I didn't imagine that we should down-cast our typed API to just strings - the user of this interface would surely expect the same type from DefaultApps as AllApps?

internal/fdo.go
```diff
@@ -174,6 +174,18 @@ func fdoLookupApplicationsMatching(appName string) []desktop.AppData {
 	return icons
 }
 
+func fdoLookupApplicationAll() []desktop.AppData {
```
andydotxyz: This uses the singlar matching fdoLookupApplicationWinInfo rather than the plural in fdoLookupApplicationsMatching which implies it's grouped with application info rather than groups of applications. Perhaps "fdoLookupApplications" is better?

icons.go
```diff
@@ -12,8 +12,10 @@ type AppData interface {
 
 // ApplicationProvider describes a type that can locate icons and applications for the current system
 type ApplicationProvider interface {
+	AllApps() []AppData
+	AvailableThemes() []string
```
andydotxyz: Why the difference between AllApps but AvailableThemes? it feels like this should be consistent.

internal/fdo.go
```diff
@@ -419,16 +461,16 @@ func findOneAppFromNames(f desktop.ApplicationProvider, names ...string) desktop
 	return nil
 }
 
-func appendAppIfExists(apps []desktop.AppData, app desktop.AppData) []desktop.AppData {
```
andydotxyz: As mentioned above I'm pretty sure that this chunk should revert to the more complete API listing app data.

settings.go
```diff
@@ -1,46 +1,201 @@
...
 
 import (
+	wmtheme "fyne.io/desktop/theme"
+	"fyne.io/fyne/canvas"
+	"fyne.io/fyne/cmd/fyne_settings/settings"
+	"fyne.io/fyne/layout"
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
 	"os"
+	"os/exec"
```
andydotxyz: These imports are not ordered correctly - system packages before imports

settings.go
```diff
@@ -1,46 +1,201 @@
...
 	}
 
-	env = os.Getenv("FYNEDESK_BACKGROUND")
+	env = os.Getenv("FYNEDESK_ICONTHEME")
 	if env != "" {
-		d.background = env
+		d.iconTheme = env
 	} else {
-		d.background = fyne.CurrentApp().Preferences().String("background")
+		d.iconTheme = fyne.CurrentApp().Preferences().String("icontheme")
```
andydotxyz: Should we not fall back to "hicolor" as before if there is no setting specified?

desk.go
okratitan: Because we want to make sure that an Instance exists before calling Settings now since we depend on the Instance in Settings.

andydotxyz: But the contents of NewDeskSettings() has not changed...

andydotxyz:
Nice improved UI.
However looking at this made me realise that "Default Icons" does not really make sense for the user - it's not default at all, it's "Launcher Icons" or even just "Icons"?...

I think that the API probably should adapt too - the default referring to system defaults and a slightly different name to the configuration.

Thoughts?

desk.go
okratitan: NewDeskSettings calls settings.load() ... settings.load() need to be able to use Instance() to get the fallback default icons

background.go
```diff
@@ -9,6 +9,17 @@ import (
 	wmtheme "fyne.io/desktop/theme"
 )
 
+func updateBackgroundPath(background *canvas.Image) {
```
andydotxyz: This would be better encapsulated (and either to test) if it took the path as a parameter.
Also it would make the updateBackgrounds() call more efficient as it could call wallpaperPath() just once.

Also why the inconsistency between background and wallpaper?

bar.go
```diff
@@ -83,14 +84,49 @@ func (b *bar) WindowRemoved(win Window) {
...
+		appBar.append(obj)
 	}
-	for _, name := range desk.Settings().DefaultApps() {
-		app := desk.IconProvider().FindAppFromName(name)
+}
+
+func (b *bar) updateIcons() {
+	iconTheme := b.desk.Settings().IconTheme()
+	for _, icon := range b.icons {
+		iconRes := icon.appData.Icon(iconTheme, int(float32(b.iconSize)*b.iconScale))
```
andydotxyz: Isn't this exact code repeated in other places?

bar.go
```diff
@@ -83,14 +84,49 @@ func (b *bar) WindowRemoved(win Window) {
...
+					iconRes = wmTheme.BrokenImageIcon
+				}
+			}
+		}
+		icon.resource = iconRes
+		icon.Refresh()
+	}
+}
+
+func (b *bar) appendDefaultIcons() {
```
andydotxyz: As in previous review I wonder if "default" is really correct?

okratitan:
Okay hopefully I resolved all of that :)

andydotxyz:
Running this version changed my launcher icons to 1 missing icon and reset my background to solid grey.

okratitan:
Really?  I wonder what's different?  I can't reproduce it.

okratitan:
Regarding the icons being gone - I'm not surprised there as the setting changed from defaultapps to launchericons ... you should be able to go in and edit that by hand to get your icon config back.  Literally have no clue about your background... unless your background is not a legit path?  I guess I should check for that?

andydotxyz:
Surely there is a better fallback to having no setting than showing no icons?
An empty config should show the default icon set.

If I restart the background shows then it goes away when I apply with an empty path. (see screenshot above).

andydotxyz:
remove your config, restart, you will see no icons. This is not a good first run experience.
We can't insist that people run the settings UI to set up icons before their bar has anything in it.

andydotxyz:
As recent testing has shown the default values are not working well.
This is really easy to test and needs to be there to automate the checks for first use experience. I suggest that at a minimum unit tests create and load new configurations and validate the default bar and background are used. Then some incremental checks where possible to see that they update appropriately with settings change.

andydotxyz:
Reviewing #15 I see that this PR only covers half of the items that were listed as configurable in the issue specification. Do you want to split out the functionality not covered to new issues - or address before release?

andydotxyz:
The functionality that is in this PR works well, thanks. But it's still not enough to close #15 and has no tests.

andydotxyz:
Travis passed, not sure why this didn't return

desk_test.go
```diff
@@ -1,7 +1,11 @@
 package desktop
 
 import (
+	wmTheme "fyne.io/desktop/theme"
```
andydotxyz: these imports are out of order

desk_test.go
```diff
@@ -106,23 +123,83 @@ func (tsp testScreensProvider) ScreenForWindow(win Window) *Screen {
...
+func newTestAppProvider() ApplicationProvider {
+	return &testAppProvider{}
+}
+
 func newTestScreensProvider() ScreenList {
 	return &testScreensProvider{}
 }
 
-func TestDeskLayout_Layout(t *testing.T) {
+func NewTestDesktop() Desktop {
```
andydotxyz: No need to export a test constructor.
You'd avoid a lot of casting if this returned *deskLayout.

andydotxyz:
Again travis has passed

PR #55, opened 2019-12-18
Maximize on start

okratitan:
This fixes applications where the the maximized state and the geometry of the window were not matching.


andydotxyz:
I'm not sure why this completely rewrites the vendor folder.
Can you PR with just the required changes in it?

wm/frame.go
```diff
@@ -563,9 +571,13 @@ func newFrame(c *client) *frame {
...
 	framed.width = w
 	framed.height = h
-	framed.childWidth = attrs.Width
-	framed.childHeight = attrs.Height
-
+	if full || !decorated {
+		framed.childWidth = w
+		framed.childHeight = h
+	} else {
+		framed.childWidth = w - uint16(framed.borderWidth()*2)
```
andydotxyz: Can you confirm that the previous use of attrs.Width is no longer required here?

wm/screens.go
```diff
@@ -63,6 +63,24 @@ func (xsp *x11ScreensProvider) ScreenForWindow(win desktop.Window) *desktop.Scre
 	return xsp.screens[0]
 }
 
+func (xsp *x11ScreensProvider) ScreenForGeometry(x int, y int, width int, height int) *desktop.Screen {
+	if len(xsp.screens) > 1 {
```
andydotxyz: This logic is backwards I think, the quick exit is len(xsp.screens) == 1, so the rest of the function can be less indented.

wm/frame.go
okratitan: yes ... attrs.width and attrs.height is not always accurate... i.e. in the case of goland the w/h reported by attrs isn't maximized width and height though it is reporting it wants to be maximized... so basing child width/height off atttrs.width and attrs.height was not accurate.  The more accurate way is to take the width and height we determine for the frame minus the borders/titlebars/decorations.

wm/screens.go
okratitan: Fixed

wm/screens.go
```diff
@@ -46,36 +46,38 @@ func (xsp *x11ScreensProvider) Scale() float32 {
 }
 
 func (xsp *x11ScreensProvider) ScreenForWindow(win desktop.Window) *desktop.Screen {
-	if len(xsp.screens) > 1 {
```
andydotxyz: Haha, when you made this change I realised - this function should just call ScreenForGeometry with the window geometry surely?

wm/screens.go
```diff
@@ -46,36 +46,38 @@ func (xsp *x11ScreensProvider) Scale() float32 {
...
+		if middleW >= xx && middleH >= yy &&
+			middleW <= xx+ww && middleH <= yy+hh {
+			return xsp.screens[i]
 		}
 	}
 	return xsp.screens[0]
 }
 
 func (xsp *x11ScreensProvider) ScreenForGeometry(x int, y int, width int, height int) *desktop.Screen {
-	if len(xsp.screens) > 1 {
```
andydotxyz: Also I realise - this has no tests, should be pretty easy to fix?

okratitan: Yeah - we can create some fake screens with fake geometry etc... and test the screens functionality.  We really need to improve coverage across the board methinks.

andydotxyz: agreed. Probably a fair bit of refactoring to do after release too.
With the duplication removed this is not so much in need of new tests ;)

PR #56, opened 2019-12-18
Notify the desktop when the cursor mouses out to a window

### LINKED ISSUE #53: If mouse out from app bar goes straight to a window, mouse out is not sent to the bar
Thankfully we are already getting a pointer entry signal on windows in the window manager - we just need to send a message to the desktop that the cursor is on a window so the app bar (or in the future any widgets placed on the desktop/widget panel) can adjust accordingly.  Should be super simple and easy.

okratitan:
This Fixes #53


andydotxyz:
It just occurred to me (sorry for the delay) but should the Fyne desktop.MouseOut() not work for this?

andydotxyz:
I don't think this should be in the public API of Desktop.
Is there some way where it can be triggered without this?

okratitan:
I'm open to suggestions but I can't really think of any.  As far as I can tell the only thing wm currently accesses is the Desktop interface, not the deskLayout itself of course.  That makes this difficult to think of any other way to do this?  And now that you mention it - I think we also need a MouseIn() as I now see the bar getting the mouse out is working but doesn't get a mouse in if cursor goes straight from window to bar.  I'll add that as well - but again I'm up to hearing ideas on a better way.

okratitan:
Ah actually we don't need MouseIn() it was missing the mouse in because we don't need to send the MouseOut if the enter notify is on the root win :)

andydotxyz:
But it does not have to be in the Desktop interface to be able to access it through Instance().
For example something in the internal package that has a similar function signature - then the line of code could check if the instance is of type "RootCanvas" (for example).

desk.go
```diff
@@ -201,6 +201,22 @@ func (l *deskLayout) scaleVars(scale float32) []string {
 	}
 }
 
+// MouseInNotify can be called by the window manager to alert the desktop that the cursor has entered the canvas
+func (l *deskLayout) MouseInNotify(x int, y int) {
+	barX, barY := appBar.Position().X, appBar.Position().Y
+	barWidth, barHeight := appBar.Size().Width, appBar.Size().Height
+	if x >= barX && x <= barX+barWidth {
```
andydotxyz: I have verified that the canvas scale != 1 does in fact break this (it is comparing screen coordinates with fyne canvas positions).
My suggestion would be that as this API is outside of the wm package it should take a fyne.Position. Then the callers of this would have to scale appropriately.

desk.go
```diff
@@ -201,6 +201,22 @@ func (l *deskLayout) scaleVars(scale float32) []string {
...
+		if y >= barY && y <= barY+barHeight {
+			appBar.MouseIn(nil)
```
andydotxyz: This seems dangerous - the contract for MouseIn is that it takes a *fyne.MouseEvent, so why not construct one with the information we have? there is no button or modifier so we can just pass the position and absolute position using the position passed into this function (when updated)?

desk_test.go
```diff
@@ -44,6 +44,10 @@ func (*testDesk) Screens() ScreenList {
 	return nil
 }
 
+func (*testDesk) MouseOutNotify() {
+	return
```
andydotxyz: Surely this line is redundant?

internal/notify/mouse.go
```diff
@@ -0,0 +1,6 @@
+package notify
+
+type MouseNotify interface {
+	MouseInNotify(x int, y int)
```
andydotxyz: As mentioned above, let's use fyne.Position as it's really just the wm package that uses x, y ints - this will help to differentiate between scaled and unscaled.

wm/desk.go
```diff
@@ -4,6 +4,7 @@ package wm // import "fyne.io/desktop/wm"
 
 import (
 	"errors"
+	"fyne.io/desktop/internal/notify"
```
andydotxyz: These imports are not sorted

wm/desk.go
```diff
@@ -235,6 +236,13 @@ func (x *x11WM) runLoop() {
 			if err != nil {
 				fyne.LogError("Set Cursor Error", err)
 			}
+			if mouseNotify, ok := desktop.Instance().(notify.MouseNotify); ok {
+				mouseNotify.MouseOutNotify()
+			}
+		case xproto.LeaveNotifyEvent:
+			if mouseNotify, ok := desktop.Instance().(notify.MouseNotify); ok {
+				mouseNotify.MouseInNotify(int(ev.RootX), int(ev.RootY))
```
andydotxyz: Proably a good spot to do the scale calculations?

wm/desk.go
```diff
@@ -241,7 +240,8 @@ func (x *x11WM) runLoop() {
 			}
 		case xproto.LeaveNotifyEvent:
 			if mouseNotify, ok := desktop.Instance().(notify.MouseNotify); ok {
-				mouseNotify.MouseInNotify(int(ev.RootX), int(ev.RootY))
+				mouseNotify.MouseInNotify(fyne.NewPos(int(float32(ev.RootX)/fyne.CurrentApp().Settings().Scale()),
```
andydotxyz: This will not work with auto scale - the settings just stores the preference not the canvas scale

andydotxyz:
Looks like travis passed. Maybe it's these conflicts that are blocking the builds atm?

PR #57, opened 2019-12-18
Disable battery and brightness where necessary (i.e. desktop boxes)

### LINKED ISSUE #50: Disable Battery and Brightness
Don't show the battery and brightness progressbars if the underlying mechanisms are not available - and don't query their value.

okratitan:
Fixes #50


widgetpanel.go
```diff
@@ -149,48 +150,51 @@ func formattedDate() string {
...
 func (w *widgetPanel) createBattery() {
-	w.battery = widget.NewProgressBar()
-	w.brightness = widget.NewProgressBar()
-
-	go w.batteryTick()
+	if _, err := battery(); err == nil {
+		w.battery = widget.NewProgressBar()
+		go w.batteryTick()
+	}
+	if _, err := battery(); err == nil {
```
andydotxyz: This looks like a pretty bad typo - brightness() not battery()

widgetpanel.go
```diff
@@ -259,16 +263,21 @@ func (w *widgetPanel) CreateRenderer() fyne.WidgetRenderer {
...
-	less := widget.NewButtonWithIcon("", theme.ContentRemoveIcon(), func() {
-		w.setBrightness(-5)
-	})
-	more := widget.NewButtonWithIcon("", theme.ContentAddIcon(), func() {
-		w.setBrightness(5)
-	})
-	bright := fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, nil, less, more),
-		less, w.brightness, more)
+	var brightnessIcon, bright, batteryIcon fyne.CanvasObject
+	if _, err := battery(); err == nil {
```
andydotxyz: This does not need to be separate from the "if" later when the layout is done does it?
All the vars listed above seem superfluous when you consider that they could be added later when the layout is defined...

widgetpanel.go
```diff
@@ -202,7 +206,7 @@ func (w *widgetPanel) setBrightness(diff int) {
 	if err != nil {
 		log.Println("Error running xbacklight", err)
 	} else {
-		w.brightness.SetValue(brightness())
+		w.brightness.SetValue(floatVal)
```
andydotxyz: I tested this and every time the plus or minus buttons is pressed the change is made but the old value is displayed.
I think the reason is that `value` should be used here, not `floatVal`.
Perhaps value should be `newValue` or something to make this clearer?

PR #62, opened 2019-12-22
Transient fix, lint fix, and configure notify fix

okratitan:
A few bug fixes for release -

We only want to manage certain windows regardless of whether have the transient property, so the transient check was useless and actually messed up a few menus that didn't have transient set but had window type menu set.

Fix lint warning on MouseNotify interface.

Borderless windows need the synthetic configure notify as well decorated ones.  Refactoring out the synthetic notify to be in the release function of frame broke borderless windows getting updated correctly.  This fixes that.

### DIFF
--- a/internal/notify/mouse.go
+++ b/internal/notify/mouse.go
@@ -2,7 +2,7 @@ package notify
 
 import "fyne.io/fyne"
 
-// MouseNotify is an interface that lets separate packages like wm tell the desktop that the cursor has moved to or from the desktop canvas.
+// MouseNotify is an interface that can be used by objects interested in when the mouse enters or exits the desktop
 type MouseNotify interface {
 	MouseInNotify(fyne.Position)
 	MouseOutNotify()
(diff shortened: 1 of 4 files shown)


wm/desk.go
```diff
@@ -209,6 +209,13 @@ func (x *x11WM) runLoop() {
 				if c.(*client).id == ev.Event {
 					if x.moveResizing {
 						x.moveResizeEnd()
+						// ensure menus etc update
```
andydotxyz: This is doubling up the previous fix - should remove the other fix.
And would it not be neater to put this inside moveResizeEnd rather than in this runloop, which is already quite long.

internal/notify/mouse.go
```diff
@@ -2,6 +2,7 @@ package notify
 
 import "fyne.io/fyne"
 
+// MouseNotify is an interface that lets separate packages like wm tell the desktop that the cursor has moved to or from the desktop canvas.
```
andydotxyz: Not sure about this - it has no connection to the window manger, it's just an interface?
How about:
"MouseNotify is use to mark an object that is interested in when the mouse enters or exists the desktop"

PR #66, opened 2019-12-23
Add key bindings for app launcher

andydotxyz:
This adds a "selected app" in the list as well as the keyboard bindings:
* Alt-space to open launcher
* Escape to close it
* Enter will launch the selected app
* Up and down will change the app selected

Also new tests are added which is easier with the refactored code.


okratitan:
Looks good!

PR #67, opened 2019-12-29
Add a handy add/remove menu

andydotxyz:
For items in the bar offer to remove them, if not then offer to pin them to the launcher.

### DIFF
--- a/internal/fdo.go
+++ b/internal/fdo.go
@@ -186,25 +186,6 @@ func fdoLookupApplications() []desktop.AppData {
 	return icons
 }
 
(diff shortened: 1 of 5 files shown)


andydotxyz:
Good feedback, thanks.
Updated to add "appWindow" which holds window info and gains access to app where possible. Removed the automatic "fake" app return when one could not be found.

I think this fixes the feedback, works here - when the window has no app the menu does not appear.

andydotxyz:
I see now this depends on your installed apps. It is finding some random app that may not be working for a number of reasons (like it’s the gnome app panel?).

The only reasonable fix is to change the wininfo app lookup to check only command line...

PR #71, opened 2020-01-03
Feature/live updates

okratitan:
This will live update fynedesk on scale changes as well as any randr changes such as resolution, plugging or removing screens, orientations, etc...


andydotxyz:
As discussed on slack this works well for borders but the backgrounds don't always work when going from a large scale to a smaller one.

okratitan:
This PR has been replaced by the new multi root PR #78

PR #74, opened 2020-01-11
Update app bar settings screen - down to a single tab

andydotxyz:
Updated the app bar tab in settings to be in a single tab with a rather more intuitive UI


andydotxyz:
Merged as it's been over 2 weeks

PR #82, opened 2020-03-17
Colour borders based on border status

andydotxyz:
Liven things up a bit :)
Added a cache of the focus state as well, seemed like a small optimisation that could help


andydotxyz:
Screenshots of the changed border colouring:
fyne-borders
fyne-borders-dark

PR #84, opened 2020-03-20
Feature/dnd

okratitan:
These are fixes that make DND work in Fynedesk - So far as I can tell this works for everything except for Chrome tabs... it works for Firefox tabs even...  Still debugging Chrome tabs... but dragging text between applications, arandr, firefox tabs, etc..., etc... all work now.

### DIFF
--- a/wm/client.go
+++ b/wm/client.go
@@ -123,7 +123,11 @@ func (c *client) Focus() {
 }
 
 func (c *client) Focused() bool {
-	return focusedWin == c.win
+	active, err := windowActiveGet(c.wm.x)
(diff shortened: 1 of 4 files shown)


andydotxyz:
I think this breaks showing centred, bordered windows.
See that alt-tab still works but alt-space does not...

okratitan:
> One of these commits seems to break showing centered, bordered windows.

My apologies.  Fixed.

andydotxyz:
Thanks.
Travis is having trouble on macOS, so I'll override that check for now.

Issue #86, opened 2020-03-21
Support fixed-sized windows


andydotxyz:
When the minimum and maximum size hints are the same the window should not be resizable.
We should not update mouse cursor etc either.

PR #87, opened 2020-03-24
Size Fixes

### LINKED ISSUE #85: Windows can be resized below their minsize
We should respect the minsize hint and not allow sizing below that value.

okratitan:
This fixes size constraints such as min size, max size, and fixed size when resizing windows.

Fixes #85 and #86


andydotxyz:
Nope, still failing on mac

andydotxyz:
Other than the build failure this looks good to go

Issue #70, opened 2019-12-31
VLC opens lots of windows, contents largely garbage


andydotxyz:
When opening a video in VLC we see many windows loaded (as you can see in the dock) and most of these contain garbage.

The first window opened is the main one and contains the menu - quitting from here closes all the windows.
Unfortunately the video cannot be viewed in any of the windows, despite the sound playing.
![Screenshot from 2019-12-31 16-12-29](https://user-images.githubusercontent.com/294436/71627128-beb28900-2be8-11ea-9549-7f79821158b2.png)

andydotxyz:
The main window is now working correctly. Some graphical glitches seem to relate to the Qt toolkit, I will open another issue.

PR #89, opened 2020-04-04
This fixes several issues related to colormaps and client map requests

okratitan:
Initially we had turned SubstructureNotify on for the root window - This is incorrect as we were picking up things like map request events from clients that were not managed by us and shouldn't be managed by us and thus managing those clients incorrectly.

Introduces transient handling and colormap handling.


andydotxyz:
OK merged as-is and picked in to 01.2. The code seems to run and videos can be played. 
A big improvement over 0.1.1 - and fixes #70. Further work to fix the menu paint/scale issues may have to wait.

PR #91, opened 2020-04-10
Fix QT App corruption

andydotxyz:
Turns out QT_SCALE_FACTOR is not reliable below 1.

This ensures value is neer below 1

https://www.reddit.com/r/kde/comments/a54rul/how_to_set_scaling_to_below_100/
> Qt itself doesn't work with scale factors < 1. Try QT_SCALE_FACTOR=0.75 kcalc.


okratitan:
This looks great.  Awesome find!

PR #93, opened 2020-04-19
Feature/moduleconfig

andydotxyz:
Allow users to turn modules on and off.
Remember the setting in preferences.
Add and remove modules as required if the configuration changes.

desk-momdules


okratitan:
Looks good!!! Excited to see module support expanding.

PR #94, opened 2020-04-20
Small refactoring in main desk packages

andydotxyz:
* Make it so that WindowManager() never returns nil. Make code a little cleaner
* Change type deskLayout to desktop, better reflects it's purpose now.


andydotxyz:
Required for the full refactor, so merging - it's been 3 days anyway

PR #100, opened 2020-04-27
Add some logic for window positioning on active display

andydotxyz:
1) Track active screen when active window changes or when the window moves screen

(To follow)
2) Move windows with no speified location to a sensible location on active display
3) hook in some non-x logic for where we want new windows to appear


andydotxyz:
Apologies for the hugeness of the refactor in the middle of this. The rest of the changes are relatively small. Just popped a final small change on this as I noticed that there were some typos in refactor that broke borderless resize.

andydotxyz:
Killed in favour of #102

PR #102, opened 2020-04-30
Move window positioning logic out of X11

andydotxyz:
1. Track active screen when active window changes or when the window moves screen
2. Move windows with no specified location to a visible location on active display
3. Hook in some non-x logic for where we want new windows to appear


wm/position.go
```diff
@@ -0,0 +1,20 @@
...
+
+import (
+	"fyne.io/fynedesk"
+	"fyne.io/fynedesk/theme"
+)
+
+// PositionForNewWindow returns the suggested position for a new window of the given geometry.
+// The screen list hints at available space, but normally list.Active() is the best.
+func PositionForNewWindow(x, y int, w, h uint, screens fynedesk.ScreenList) (int, int, uint, uint) {
+	if x != 0 && y != 0 {
```
okratitan, line 11: Instead of this check - lets just test for a specified position like so:
hints, err := icccm.WmNormalHintsGet(c.wm.X(), c.win)
if err == nil {
    if hints.Flags & icccm.SizeHintPPosition > 0 && hints.Flags & icccm.SizeHintUSPosition > 0 {
        *iterate screens and check to make sure x,y instersect somewhere and isn't broken*
        if !broken { return } else we will continue on and let the top left positioning happen.
    }
}

wm/position.go
```diff
@@ -0,0 +1,20 @@
...
+		return x, y, w, h
+	}
+
+	target := screens.Active()
```
okratitan, line 15: Here let's go ahead and do something like this:
winType = windowTypeGet(c.wm.X(), win)
transient = false
transientFor = x11.WindowTransientForGet(c.wm.X(), win)
if transientFor > 0 && transientFor != win {
     transient = true
}
if winType == windowTypeNormal && !transient {
     * We know this is a top level window and place top left *
} else if winType == windowTypeDialog && transient {
     * center win on transientFor *
} else {
     * try and use provided x, y geometry (and fix if broken) because this could be a dnd window, menu, utility that needs special placement, etc...
}

andydotxyz: Hmm. We don't expose transientFor properties through the API currently... or type for that matter.
Could that be added in a later PR when we figure how to write that in a non-X specific way?

wm/position.go
andydotxyz: I have found a way of doing this I think

andydotxyz: This is fixed but in the X code as those properties are not exposed. This is now called conditionally - if it was not positioned or if the requested position was bad

PR #104, opened 2020-05-04
Feature/sound widget

mattboll:


andydotxyz:
Do you have a screenshot of the updated work? I'm not on Linux today so can't test sorry

mattboll:
sure, right now it looks like this
screeshot0510
screeshot0510_mute

andydotxyz:
Nice work thanks.
It works when my laptop is on it's own - but when I plug it into my doc it stops working. I think that I have to look at my sound configuration.

mattboll:
can you run pavucontrol and see what is the default output ?

The module should only change the current default output, if for any reason the current default output is not where the sound comes from, then it won't be changed. May be all the outputs should have the same volume but I'm not sure it's a good idea

mattboll:
Like I said by slack, it should mostly work, please tell me what would be the minimal things you would like to allow merge and I'll be happy to complete this PR or open an other PR

andydotxyz:
It looks like the application (Chromium) was playing to the wrong channel so maybe it was all working?

mattboll:
It works as expected then but we can add some improvements, we will need to define what we want :)

andydotxyz:
It looks like your patches to the pulseaudio lib were accepted, should this be updated to use upstream?

go.sum
```diff
@@ -8,15 +8,26 @@ github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802 h1:1BDTz0u9nC3//pOC
...
+github.com/erikstmartin/go-testdb v0.0.0-20160219214506-8d10e4a1bae5/go.mod h1:a2zkGnVExMxdzMo3M0Hi/3sEU+cWnZpSni0O6/Yb/P0=
+github.com/ernesto-jimenez/gogen v0.0.0-20180125220232-d7d4131e6607/go.mod h1:Cg4fM0vhYWOZdgM7RIOSTRNIc8/VT7CXClC3Ni86lu4=
+github.com/fatih/structs v1.0.0/go.mod h1:9NiDSp5zOcgEDl+j00MP/WkGVPOlPRLejGD8Ga6PJ7M=
 github.com/fsnotify/fsnotify v1.4.9 h1:hsms1Qyu0jgnwNXIxa+/V/PDsU6CfLf6CNO8H7IWoS4=
 github.com/fsnotify/fsnotify v1.4.9/go.mod h1:znqG4EE+3YCdAaPaxE2ZRY/06pZUdp0tY4IgpuI1SZQ=
 github.com/fyne-io/mobile v0.0.2-0.20200421111809-d9d371319719 h1:eRE3ElW0LlpTfyPp0GvZdXIIzBbfCuFtFJkqTRzWryQ=
 github.com/fyne-io/mobile v0.0.2-0.20200421111809-d9d371319719/go.mod h1:/kOrWrZB6sasLbEy2JIvr4arEzQTXBTZGb3Y96yWbHY=
 github.com/fyne-io/xgbutil v0.0.0-20191220152344-7d838166824d h1:8+poQtDqiRfAZmeyYRlInuaIBsQEflmxtpJeZklvNuE=
 github.com/fyne-io/xgbutil v0.0.0-20191220152344-7d838166824d/go.mod h1:vPqYMuTmTyrLNDQWQXy9lqdnV+WRQn6jhaSon8TppBs=
+github.com/ghetzel/go-defaults v1.2.0/go.mod h1:xWhTgOoc4UNWT7sl3oyNFqtKzEbUKI9C3rqbwtwdxFw=
```
andydotxyz, line 27: Does this change require all these additional dependencies?
Have you run `go mod tidy` recently?

modules/builtin/pulseaudio.go
```diff
@@ -0,0 +1,110 @@
...
+	"fyne.io/fyne/theme"
+	"fyne.io/fyne/widget"
+	"github.com/mattboll/pulseaudio"
+
+	"fyne.io/fynedesk"
+	wmtheme "fyne.io/fynedesk/theme"
+)
+
+var pulseaudioMeta = fynedesk.ModuleMetadata{
+	Name:        "Pulseaudio",
```
andydotxyz, line 17: This string (and the overall naming) I think would be better as "Audio" or "Sound" - it is specific to pulse just now, but over time this module would surely manage all audio for fynedesk?

modules/builtin/pulseaudio.go
```diff
@@ -0,0 +1,110 @@
...
+	NewInstance: NewPulseaudio,
+}
+
+// Pulseaudio is a progress bar module to modify screen pulseaudio
+type Pulseaudio struct {
+	bar       *widget.ProgressBar
+	client    *pulseaudio.Client
+	soundIcon *widget.Button
```
andydotxyz, line 25: this name seems to conflict purpose - icon != button :) Maybe "mute" is suitable now?

modules/builtin/pulseaudio.go
```diff
@@ -0,0 +1,110 @@
...
+	"fyne.io/fynedesk"
+	wmtheme "fyne.io/fynedesk/theme"
+)
+
+var pulseaudioMeta = fynedesk.ModuleMetadata{
+	Name:        "Pulseaudio",
+	NewInstance: NewPulseaudio,
+}
+
+// Pulseaudio is a progress bar module to modify screen pulseaudio
```
andydotxyz, line 21: I suspect that "modify screen pulseaudio" is a copy-paste bug :-p

modules/builtin/pulseaudio.go
```diff
@@ -0,0 +1,110 @@
...
+	client    *pulseaudio.Client
+	soundIcon *widget.Button
+}
+
+// Destroy destroys the module
+func (b *Pulseaudio) Destroy() {
+	b.client.Close()
+}
+
+func (b *Pulseaudio) value() (float32, bool, error) {
```
andydotxyz, line 33: I don't think the bool param is ever used

modules/builtin/pulseaudio.go
```diff
@@ -0,0 +1,110 @@
...
+	if err != nil {
+		return 0, true, err
+	}
+
+	return volume, false, nil
+}
+
+// OffsetValue actually increase or decrease the screen pulseaudio
+func (b *Pulseaudio) OffsetValue(diff int) {
+	floatVal, _, _ := b.value()
```
andydotxyz, line 44: Error is silently ignored...

go.sum
mattboll: I didn't know about go mod tidy thanks I was struggling about ghetzel, I didn't understand why it always came back

modules/builtin/pulseaudio.go
mattboll: Pulseaudio plugin and alsa plugin could be slightly different, I'm not sure yet but it's possible that alsa plugin can't change sound by application, ok right now it's not possible with the pulseaudio widget too but I hope that will change
I also looked at xfce which has both plugins and not one sound plugin
That's why I named Pulseaudio but I can change my mind

modules/builtin/pulseaudio.go
mattboll: /o\ shame on me !

modules/builtin/pulseaudio.go
mattboll: Yes good spot, due to refactor the mute value doesn't come from this function anymore

mattboll:
Thanks for your time @andydotxyz , I appreciate your comments, it should be better now

internal/ui/settings.go
```diff
@@ -179,7 +179,7 @@ func (d *deskSettings) load() {
 		d.launcherZoomScale = 2.0
 	}
 
-	moduleNames := fyne.CurrentApp().Preferences().StringWithFallback("modulenames", "Battery|Brightness")
+	moduleNames := fyne.CurrentApp().Preferences().StringWithFallback("modulenames", "Battery|Brightness|Pulseaudio")
```
andydotxyz, line 182: I think this line won't load the module any more?

mattboll: of course, I didn't remember that /o\

PR #105, opened 2020-05-05
Support (configurable) key bindings for modules etc

### LINKED ISSUE #37: Configurable Desktop Shortcuts
There should be a way to define shortcuts to perform common tasks, such as bringing up a launcher or terminal, clearing notifications or taking a screenshot

andydotxyz:
This is a work in progress towards proper key bindings.

- [x] Started with allowing code to declare shortcuts...
- [x] Moving through modules so they can declare shortcuts.
- [x] Added a settings UI to list shortcuts

TODO

- [ ] Reset bindings when modules are loaded / unloaded
- [ ] Make these bindings configurable in settings

Will fix #37


andydotxyz:
Maybe @mattbol could look at this and see if they agree with the changes to brightness key binding...

andydotxyz:
Maybe @mattboll could look at this and see if they agree with the changes to brightness key binding...

mattboll:
sure I can throw an eye, the code looks great, hopefully I'll be able to test tonight

mattboll:
I'm not sure that alt+shift+tab works as expected but I'm not sure that it was different on develop… anyway, brightness works as expected with keyboard shortcut :+1:

andydotxyz:
Can you say what wasn’t as expected? It looks ok to me

mattboll:
Imagine I've got Firefox, Intellij, Terminal, Slack, in this order and looking at firefox
When I do Alt-Tab (and keep my finger on alt) I switch to intellij, then I tap on tab and switch to terminal, then i tap on tab and switch to slack and so on…
If I do Alt-Shift-Tab I expect to go first to slack, then terminal, then intellij (reverse order from just tab)

Also, if I do Alt-Tab (keeping finger on Alt) and then tap on Shift-Tab, I should return to my first window.

Right now, I see no difference between alt-tab and alt-shift-tab

But having a better look at the patch I can see that the call of `x.showOrSelectAppSwitcher(shiftPressed)` function hasn't change

andydotxyz:
I will have to look into this further because as far as I can tell it does work as you describe here... maybe I have some local state, will try resetting.

andydotxyz:
OK, I think this is ready for review

andydotxyz:
Cool, I will open a new PR for the configuration options later

PR #109, opened 2020-05-11
WIP - Initial install script using GNU Make

Jacalz:
This currently requires that the user runs `make install DESTDIR=/usr` or `make install DESTDIR=/usr/local`, but we might want to get that automated.
There are also no errors and exits unlike the old install.sh script.


mattboll:
Also, almost everybody needs sudo or root to make install but we shouldn't use root to build. Tell me if i'm wrong but I think that right now if you run make build as user, then sudo make install, it will re-build with sudo rights

Jacalz:
Thanks for the comments @mattboll. I have now fix the issues that you mentioned and added much needed automation for /usr/local or /usr as DESTDIR.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+else
+DESTDIR ?= /usr
+endif
+
+build:
+	go build ./cmd/fynedesk_runner || (echo "Failed to build fynedesk_runner"; exit 1)
+	go build ./cmd/fynedesk || (echo "Failed to build fynedesk"; exit 1)
+
+install:
+	cp cmd/fynedesk_runner/fynedesk_runner $(DESTDIR)$(PREFIX)/bin
```
andydotxyz, line 13: I don't think prefix belongs here any more

README.md
```diff
@@ -21,22 +21,27 @@ any contributions or feedback for the project.
...
-Using standard go tools you can install Fyne's desktop using:
+Using standard go tools you can install FyneDesk using:
+```
+go get fyne.io/fynedesk/cmd/fynedesk
+```
 
-    go get fyne.io/fynedesk/cmd/fynedesk
+Once installed you can run the following commands to get everything set up. This will take care of all the building and configuration for you.
+```
+make
```
andydotxyz, line 36: Before this works you need to be in the right directory, which the user will not be if they just executed 'go get'.

README.md
```diff
@@ -21,22 +21,27 @@ any contributions or feedback for the project.
...
+make install
```
andydotxyz, line 37: Shouldn't this be `sudo make install`?

Jacalz: You are right. Missed that for some reason.

README.md
Jacalz: Good catch. Didn't think of that.

PR #123, opened 2020-06-30
Copy log files out to crash logs when something goes wrong

andydotxyz:
makes "fynedessk-crash-<date>.log" files alongside fynedesk.log when it detects a crash.

### DIFF
--- a/cmd/fynedesk_runner/log.go
+++ b/cmd/fynedesk_runner/log.go
@@ -1,34 +1,49 @@
 package main
 
 import (
-	"io"
+	"fmt"
 	"os"
 	"path/filepath"
+	"time"
 
 	"fyne.io/fyne"
 )
 
-func logPath() string {
-	return logPathRelativeTo(systemLogDir())
+func crashLogPath() string {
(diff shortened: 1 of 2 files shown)


cmd/fynedesk_runner/log.go
```diff
@@ -1,34 +1,49 @@
...
+	return logPathRelativeTo(systemLogDir())
+}
+
+func logPathRelativeTo(parent string) string {
+	return filepath.Join(logDir(parent), "fynedesk.log")
 }
 
-// openLogWriter returns a stream for stdOut and stdErr of the process we run
-func openLogWriter() (io.WriteCloser, io.WriteCloser) {
+// openLogWriter returns the log file for stdOut and stdErr of the process we run
```
okratitan, line 40: This comment seems incorrect now that it only returns stderr.

andydotxyz: Hmm, maybe - it's not incorrect though, as both streams should be writing to it?

andydotxyz: Updated the comment, removing possible confusion

PR #126, opened 2020-07-12
Feature/launchmodule

andydotxyz:
Add a new extension point for modules - launcher suggestions.

Add url opener and calculations into suggest list (only calculations on by default).


internal/ui/settings.go
```diff
@@ -179,7 +179,7 @@ func (d *deskSettings) load() {
 		d.launcherZoomScale = 2.0
 	}
 
-	moduleNames := fyne.CurrentApp().Preferences().StringWithFallback("modulenames", "Battery|Brightness|Sound")
+	moduleNames := fyne.CurrentApp().Preferences().StringWithFallback("modulenames", "Battery|Brightness|Sound|Launch Calculations")
```
okratitan, line 182: I'm not sure I'm crazy about the naming here "Launch Calculations"

okratitan: I like "Launch Suggestions" better I think

modules/launcher/calc.go
```diff
@@ -0,0 +1,87 @@
...
+	"fyne.io/fyne"
+
+	"fyne.io/fynedesk"
+	wmTheme "fyne.io/fynedesk/theme"
+
+	"github.com/Knetic/govaluate"
+)
+
+var calcMeta = fynedesk.ModuleMetadata{
+	Name:        "Launch Calculations",
```
okratitan, line 16: Same as above... "Calculations" feels weird here.

modules/launcher/urls.go
```diff
@@ -0,0 +1,68 @@
...
+}
+
+func (r *urlResult) Title() string {
+	return r.url
+}
+
+func (r *urlResult) Launch() {
+	u, err := url.Parse(r.url)
+	if err != nil {
+		fyne.LogError("COuld not parse URL", err)
```
okratitan, line 63: capitalization error here.

modules/launcher/calc.go
okratitan: I see my mixup now. Disregard.

internal/ui/settings.go
okratitan: I see my mixup now. Disregard.

okratitan: Why is the calculator module the only one loaded here? Why not the URL as well?

okratitan, line 182: Why is the calculator launch module the only one loaded here?  Why not URL?

Also - I suppose it would be trivial to add an open in terminal with paths module?

andydotxyz: It seemed like calculator was clearly beneficial but URLs perhaps less so.
On reflection maybe they should all be in a single "Launch Suggestions" module and just remove the user ability to turn them on/off (until we add module config)?

Issue #125, opened 2020-07-03
Double click titlebar to maximize


andydotxyz:
that would be lovely please :)

Also new windows starting in maximized that pass their own size don't get true maximized if the size is wrong.

okratitan:
Merged to develop. Closing.

PR #130, opened 2020-07-21
Add the ability to maximize/unmaximize windows by double clicking titlebar

okratitan:
This is an attempt at handling double click to maximize/unmaximize using go's Context api.

Addresses #125

### DIFF
--- a/internal/x11/win/frame.go
+++ b/internal/x11/win/frame.go
@@ -3,7 +3,9 @@
 package win
 
 import (
+	"context"
 	"image"
+	"time"
 
 	"fyne.io/fyne"
 	"fyne.io/fyne/driver/desktop"
@@ -34,6 +36,10 @@ type frame struct {
 
 	borderTop, borderTopRight xproto.Pixmap
 	borderTopWidth            uint16
+
+	doubleClick bool
+	ctx context.Context
+	cancelFunc context.CancelFunc
 
 	canvas fyne.Canvas
 	client *client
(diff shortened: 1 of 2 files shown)


internal/x11/win/frame.go
```diff
@@ -579,17 +585,40 @@ func (f *frame) mouseRelease(x, y int16) {
...
 		return
 	}
-	obj.(fyne.Tappable).Tapped(&fyne.PointEvent{})
+	go func() {
+		f.ctx, f.cancelFunc = context.WithDeadline(context.TODO(), time.Now().Add(time.Millisecond*300))
+		defer f.cancelFunc()
+		select {
+		case <-f.ctx.Done():
+			if f.doubleClick == true {
+				if f.client.Maximized() {
```
andydotxyz, line 599: This is putting border logic back into the X11 package which is what I've been trying to avoid.
Instead it would be cleaner to put the double tap /handling/ into the UI, and here just trigger double tap handlers - i.e. the `fyne.DoubleTappable` interface.

internal/x11/win/frame.go
```diff
@@ -579,17 +585,40 @@ func (f *frame) mouseRelease(x, y int16) {
...
+					f.client.Unmaximize()
+				} else {
+					f.client.Maximize()
+				}
+			} else {
+				obj := wm.FindObjectAtPixelPositionMatching(int(relX), int(relY), f.canvas,
```
andydotxyz, line 605: Keeping this inline makes the code much more complex to read.
It looks like some functions should be factored out here, as there are many scopes:
*) top level - handles a mouse release
*) go routine - runs a timer to detect type of mouse release
*) first fork - has what to do if double tapped
*) second fork - the previous tap handling code

That's a lot of different scopes for one method.

internal/x11/win/frame.go
```diff
@@ -35,6 +37,10 @@ type frame struct {
 	borderTop, borderTopRight xproto.Pixmap
 	borderTopWidth            uint16
 
+	doubleClick bool
+	ctx context.Context
```
andydotxyz, line 41: This variable (and possibly others?) can be limited to the function that fires off the goroutine can it not?

p.s. `doubleClick` sounds like a verb, perhaps something like 'clickCount' or other name might better indicate it's purpose?

PR #136, opened 2020-10-08
Feature/bsd

andydotxyz:
Getting FreeBSD support up to scratch


modules/status/sound_other.go
```diff
@@ -0,0 +1,60 @@
...
+package status
+
+import (
+	"log"
+
+	wmtheme "fyne.io/fynedesk/theme"
+	"github.com/mafik/pulseaudio"
+)
+
+// Destroy destroys the module
```
okratitan, line 12: Hmm perhaps some better wording here.  "Destroy destroys" just sounds odd.  Destroy removes the module maybe?

modules/status/sound_other.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		return err
+	}
+	b.client = client
+	return nil
+}
+
+func (b *sound) setValue(vol int) {
+	if err := b.client.SetVolume(float32(vol) / 100); err != nil {
+		log.Println("Failed to set volume", err)
```
okratitan, line 40: Should we be using the fyne logging functions here?

modules/status/sound_other.go
```diff
@@ -0,0 +1,60 @@
...
+		return
+	}
+
+	b.bar.SetValue(float64(vol))
+}
+
+func (b *sound) toggleMute() {
+	toggl, err := b.client.ToggleMute()
```
okratitan, line 48: If writing out toggl why not use the actual word toggle?

modules/status/sound_other.go
```diff
@@ -0,0 +1,60 @@
...
+	if err != nil {
+		log.Println("toggleMute() failed", err)
+		return
+	}
+
+	if toggl {
+		b.mute.SetIcon(wmtheme.MuteIcon)
+	} else {
```
okratitan, line 56: perhaps could return early in the if and not have the else statement

modules/status/sound_bsd.go
```diff
@@ -0,0 +1,76 @@
...
+	level := fmt.Sprintf("%d:%d", vol, vol)
+	cmd := exec.Command("mixer", "vol", level)
+	if err := cmd.Run(); err != nil {
+		log.Println("Failed to set volume", err)
+		return
+	}
+
+	b.bar.SetValue(float64(vol))
+	if vol == 0 {
+		b.mute.SetIcon(wmtheme.MuteIcon)
```
okratitan, line 56: Could return early and not have the else

modules/status/sound_bsd.go
```diff
@@ -0,0 +1,76 @@
...
+	"os/exec"
+	"strconv"
+	"strings"
+
+	wmtheme "fyne.io/fynedesk/theme"
+)
+
+var volume int
+
+// Destroy destroys the module
```
okratitan, line 17: "Destroy destroys" seems odd.  Perhaps Destroy removes the module?

internal/icon/xpm.go
```diff
@@ -0,0 +1,125 @@
...
+
+	return img
+}
+
+func parseColor(data string) (id string, c color.Color) {
+	if len(data) == 0 {
+		return
+	}
+	parts := strings.Fields(data)
+	if len(parts) != 3 {
```
okratitan, line 52: This seems like an unnecessary if statement since the next line check if len(parts) == 2

internal/icon/xpm.go
```diff
@@ -0,0 +1,125 @@
...
+
+func parseXPM(data []byte) image.Image {
+	var colCount, charSize int
+	colors := make(map[string]color.Color)
+	var img *image.NRGBA
+
+	rowStart := 0
+	rowNum := 0
+	for i, b := range data {
+		if b == '\n' {
```
okratitan, line 19: This entire section could be unnested by changing this if to if b != '\n' { continue }

okratitan (changes requested):

modules/status/sound_other.go
andydotxyz: This was copied from all other modules, I can review them all and update accordingly

modules/status/sound_other.go
andydotxyz: Sure thing, This was in line with the other volume code, I'l update the older code too

modules/status/sound_other.go
andydotxyz: This was old code refactored, will update

modules/status/sound_bsd.go
andydotxyz: Because in this instance there are two different actions, 1 for muted and 1 for not.

modules/status/sound_other.go
andydotxyz: In this instance there are two different actions, 1 for muted and 1 for not.
Returning early is a good preference for handling edge cases when the body of the method is to come. In this instance both lines have equal importance.

modules/status/sound_bsd.go
andydotxyz: This was old code refactored, will update

internal/icon/xpm.go
andydotxyz: Nice catch thanks

internal/icon/xpm.go
andydotxyz: I don't follow. There are 3 cases handled.
* 3 parts (skips)
* 2 parts (adds the "c")
* other (probably error, just returns)

okratitan: so instead of nesting the whole if/else and the important code, just nest the return in the else
if len(parts) == 2 && parts[0] == "c" {
        parts = []string{" ", "c", parts[1]}
} else {
        if len(parts) != 3 {
	        return
        }
}

andydotxyz: As indicated there it is moved rather than simplified as there is still the double nesting.
However switch that to an "else if" and you have a point. Updating.

okratitan (approved):

PR #141, opened 2020-10-19
Feature/printscreen

andydotxyz:
Add print screen features
* PrintScreen button captures all displays
* Shift+PrintScreen captures the current window with border.

The images are generated in the windowmanager code but our UI overlay then presents it to the user and offers to save.


Jacalz (commented):
This is a great feature. I do have some issues with Fynedesk crashing when I try to take a screenshot of GTK apps like Firefox and it seems like capturing GTK apps doesn't work in full screen mode either.
firefox-empty

andydotxyz:
From your screenshot it's hard to say what should have been shown - was firefox full screen on the left window?

Jacalz:
Yeah that is a very true point. It was not full-screen, I think, but I do know that I tested Nautilus (the standard gnome file-manager) and it did not show either. Please also notice that the borders on `wormhole-gui` are gone.

andydotxyz:
@Jacalz thinking more about this the missing borders were fixed in one of the later commits in this PR - can you please double check that you were testing against the latest version of this PR?

It is possibly still an issue for other reasons, but one issue was resolved during development :)

andydotxyz:
And to confirm, when your windows are missing they are leaving those white rectangles rather than being omitted completely?

Jacalz:
@andydotxyz I updated to the latest version and it appears to still be the case that borders are not shown. They do appear as dark in the preview for the screenshot though. Perhaps they are transparent? Yup, double checked in GIMP and they are indeed transparent.
image

Yes, the gtk windows are replaced with the white rectangles. I have never seen any windows be left out entirely. Please see the attached screenshot below:
firefox-empty

Aspect ratio for the screenshot viewer seems to be wrong until it gets some sort of response (click, drag or something that triggers a refresh) and then it gets the right aspect ratio.
strange-aspect

I have also been seeing once in a while that the screenshot viewer will not show any picture after taking a screenshot. I cannot replicate it reliably but have seen it once or twice at least.

andydotxyz:
Thanks. The aspect issues were resolved in Fyne earlier today, and should be OK now.
I will try to replicate the GTK issues later tonight

andydotxyz:
Strange, here is nautilus at my end...
nautilus

andydotxyz:
Firefox worked too...
firefox

andydotxyz:
Perhaps the issue @Jacalz finds is due to screen depth not matching the colour depth on the borders?
Can you share anything about the colour confugration in your Xorg @Jacalz.

Have you been able to test this yet @okratitan - it's been sitting for a month now and I'd like to gauge whether I am lucky in this working for me, or unlucky that it doesn't work on some old configuration...

Jacalz:
@andydotxyz Sorry, but I am not entireley sure where to look for that sort of colour information. Do you perhaps know where that information is usually stored?

andydotxyz:
It might be determined at runtime - but some X summary printout or maybe xrandr could list it?

Jacalz:
Looks like I still have the same issues with windows being blank (transparent). Also does look like it can't capture the second monitor at all after the multidesk fix.

Jacalz:
fynedesk-printscreen
This is what I am seeing on the latest commit. Everything that isn't a fyne application is transparent and the second monitor is not captured at all.

internal/ui/screenshot.go
```diff
@@ -0,0 +1,70 @@
...
+	preview := canvas.NewImageFromImage(img)
+	preview.FillMode = canvas.ImageFillContain
+	w.SetContent(container.NewBorder(nil, buttons, nil, nil, preview))
+	w.Resize(fyne.NewSize(400, 250))
+	w.Show()
+}
+
+func saveImage(pix image.Image, w fyne.Window) {
+	d := dialog.NewFileSave(func(write fyne.URIWriteCloser, err error) {
+		if err != nil {
```
Jacalz, line 57: You are not checking if `write == nil`. I was noticing that pressing save and then cancelling crashes the desktop.

internal/ui/screenshot.go
```diff
@@ -0,0 +1,70 @@
...
+		layout.NewSpacer(),
+		widget.NewButton("Cancel", func() {
+			w.Close()
+		}),
+		save)
+
+	preview := canvas.NewImageFromImage(img)
+	preview.FillMode = canvas.ImageFillContain
+	w.SetContent(container.NewBorder(nil, buttons, nil, nil, preview))
+	w.Resize(fyne.NewSize(400, 250))
```
Jacalz, line 51: Could we perhaps make the window slightly larger? The window is too small to make the save dialog big enough to use.

Jacalz (changes requested):
Thanks for the latest fix. Screenshots are now working :1st_place_medal: 
Just two smaller issues, but otherwise approved :+1:

Jacalz (approved):
Looks good to me. Staticcheck is failing, so you might need to manually merge this.

PR #142, opened 2020-10-19
Clean up Makefile and check for destdir

Jacalz:
Also makes sure that binaries are ignored by git.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,8 @@
+# Ignore binaries
 cmd/fynedesk/fynedesk
 cmd/fynedesk_runner/fynedesk_runner
+fynedesk
+fynedesk_runner
 
 .idea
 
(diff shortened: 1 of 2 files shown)


andydotxyz:
Thanks for this, but you didn't describe why the fail message and return was removed - is that not a useful function?

Jacalz:
Ah, right. Sorry. As make prints out each line before the executing it, I felt like it would be much clearer what went wrong. As an added benefit that the text that is printed out is more way to read and understand.

andydotxyz:
Ah, so make will still exit early if we fail on one of those steps?

Jacalz:
Good question. I hadn't thought about that part. Will check.

Jacalz:
Looks like it indeed will do so.
image

andydotxyz (approved):

PR #144, opened 2020-10-24
Enable staticcheck and fix all issues

Jacalz:
This PR makes sure to get all the issues reported by staticcheck fixed and marks the ones that can't be fixed to be ignored for now (waiting for Fyne 2.0.0 mostly). It should now be possible to set the GitHub action to be required :+1:


internal/ui/widgetpanel.go
```diff
@@ -134,11 +134,8 @@ func (w *widgetPanel) showAccountMenu(from fyne.CanvasObject) {
 		w.desk.WindowManager().Close()
 	}))
 
-	popup := widget.NewPopUpMenu(fyne.NewMenu("Account", items...), root.Canvas())
-
 	bottomLeft := fyne.CurrentApp().Driver().AbsolutePositionForObject(from)
-	popup.Move(bottomLeft.Subtract(fyne.NewPos(0, popup.MinSize().Height)))
-	popup.Resize(fyne.NewSize(from.Size().Width, popup.MinSize().Height))
+	widget.ShowPopUpMenuAtPosition(fyne.NewMenu("Account", items...), root.Canvas(), bottomLeft)
```
Jacalz, line 138: Don't know if this is the best solution either. It is what I came up with and I think it seems to work.

Jacalz:
Staticcheck is being killed for some strange reasons. Was this what you were seeing before you turned off staticcheck? @andydotxyz

andydotxyz:
> Was this what you were seeing before you turned off staticcheck? @andydotxyz

Yes

Jacalz:
@dominikh Do you have any idea as to why this would be happening to staticcheck?

dominikh:
It's being killed because of an out of memory situation. This is probably https://github.com/dominikh/go-tools/issues/825.

internal/ui/widgetpanel.go
andydotxyz: With this change is it still the same width as the widget panel?
What was the error that this change fixes?

screen.go
```diff
@@ -29,30 +29,23 @@ type Screen struct {
 
 // CanvasScale calculates the scale for the contents of a desktop canvas on this screen
 func (s *Screen) CanvasScale() float32 {
```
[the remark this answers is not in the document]
andydotxyz: I don't think so - what error is it fixing?

screen.go
```diff
@@ -29,30 +29,26 @@ type Screen struct {
...
+			} else if scale != 0 {
+				return float32(scale)
+			}
 		}
-		fyne.LogError("Error reading scale", err)
-	}
 
-	if env != "auto" {
-		setting := fyne.CurrentApp().Settings().Scale()
-		if setting != fyne.SettingsScaleAuto && setting != 0.0 {
```
andydotxyz, line 53: Like above I think we need to check against auto - that's how older versions of Fyne stored the "auto" user preferrence that is now default

andydotxyz (changes requested):
I think one or two of my previous comments should have been "request changes" so applying that now

screen.go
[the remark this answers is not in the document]
Jacalz: Sorry for a late answer. It is trying to work around the fact that `fyne.SettingsScaleAuto` is deprecated.

internal/ui/widgetpanel.go
Jacalz: Sorry for a late answer. It is trying to work around the fact that `widget.NewPopUpMenu` is deprecated. Is this just deprecated in the meaning that it no longer opens the popup on calling it? If so, then I can happily revert this :)

andydotxyz: That is the case yes.

screen.go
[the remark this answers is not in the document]
Jacalz: Do you have any suggestions how to solve it differently?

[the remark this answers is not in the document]
andydotxyz: No. It is deprecated but if we stop using it then we lose the upgrade-path for folk who had that set in their configuration.
Either we have to copy the value (seems bad) or it needs to be marked as ignore the error

internal/ui/baricon.go
```diff
@@ -138,7 +138,7 @@ func (bi *barIcon) TappedSecondary(*fyne.PointEvent) {
 
 	c := fyne.CurrentApp().Driver().CanvasForObject(bi)
 	pos := fyne.CurrentApp().Driver().AbsolutePositionForObject(bi)
-	menu = widget.NewPopUpMenuAtPosition(fyne.NewMenu("", []*fyne.MenuItem{addRemove}...), c, pos)
+	widget.ShowPopUpMenuAtPosition(fyne.NewMenu("", []*fyne.MenuItem{addRemove}...), c, pos)
```
andydotxyz, line 141: Sorry, I think this breaks the `menu.Hide` code in the `addRemove` callback. There is no reference to the menu any more

internal/x11/win/frame.go
```diff
@@ -342,7 +342,6 @@ func (f *frame) drawDecoration(pidTop xproto.Pixmap, drawTop xproto.Gcontext, pi
 		b.SetMaximized(f.client.maximized)
 		// TODO maybe update icon?
 	}
-	f.canvas.SetScale(scale)
```
andydotxyz, line 345: This line is required

andydotxyz (changes requested):

andydotxyz:
I think that if this was rebased the checks would start running again

Jacalz:
It should now be updated against latest develop again. Going to see if I can fix these comments.

screen.go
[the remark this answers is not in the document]
Jacalz: Marked it to be ignored for now.

screen.go
Jacalz: Marked as ignore for now. We can tackle this later.

internal/ui/baricon.go
Jacalz: Good catch. It is a kind of strange way to do the code, but I have reverted this for now.

internal/x11/win/frame.go
Jacalz: It looks to be deprecated (`(345, 2)  SA1019  f.canvas.SetScale is deprecated: Settings are now calculated solely on the user configuration and system setup.`). Reverted and ignored for now.

Jacalz:
Looks like we are still seeing `staticcheck` getting killed.

andydotxyz:
So #148 would fix this maybe?

Jacalz:
Hopefully. Lets rebase on it when that one lands :slightly_smiling_face:

andydotxyz (approved):

PR #145, opened 2020-10-24
Add option to use 24h time (also fixes 12h time)

Jacalz:
This PR adds support for changing between 12h and 24h time (12h is default).

## 12h time (fixed to be true 12h):
image

## 24h time:
image

### DIFF
--- a/internal/ui/settings.go
+++ b/internal/ui/settings.go
@@ -18,6 +18,7 @@ type deskSettings struct {
 	launcherDisableTaskbar bool
 	launcherDisableZoom    bool
 	launcherZoomScale      float64
+	clockFormatting        string
 
 	moduleNames []string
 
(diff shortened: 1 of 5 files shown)


internal/ui/settings.go
```diff
@@ -183,6 +194,11 @@ func (d *deskSettings) load() {
 	if moduleNames != "" {
 		d.moduleNames = strings.Split(moduleNames, "|")
 	}
+
+	d.clockFormatting = fyne.CurrentApp().Preferences().StringWithFallback("clockformatting", "12h")
+	if d.clockFormatting != "12h" && d.clockFormatting != "24h" {
```
andydotxyz, line 199: Is this code block needed?
In widgetpanel.go there is a `if == "12h"` - if that were reversed to "24h" then this would not be needed I think... anything other than "24h" will show 12, which is the default

Jacalz: Good point. Should probably be possible to remove it.

andydotxyz (approved):

andydotxyz:
Looks great - I can't believe I never noticed 12hr was using 24 oops!

PR #146, opened 2020-10-28
Update multiple roots to only control Primary with Fyne

andydotxyz:
This speeds up load and reconfigure.
Seems more stable for the screen configuration too.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/godbus/dbus/v5 v5.0.4-0.20200513180336-df5ef3eb7cca
 	github.com/jackmordaunt/icns v1.0.1-0.20200413110149-9e181b441ab2
 	github.com/mafik/pulseaudio v0.0.0-20200511091429-8449222912dd
+	github.com/nfnt/resize v0.0.0-20180221191011-83c6a9932646
 	github.com/stretchr/testify v1.5.1
 	howett.net/plist v0.0.0-20181124034731-591f970eefbb
 )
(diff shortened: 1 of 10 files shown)


okratitan (commented):
This looks good and I am currently running this PR and it definitely feels snappy and seems to initially configure well.  However there are a ton of issues and crashes now when it comes to reconfiguring screens using something like arandr.  Resolutions go wackadoodle and can never get back to the original - clicking the widgets on the primary screens constantly crash after switching primary screen and coming back to the original. Scale is wackadoodle if primary screens are changed to a screen with different scale and back to the original etc...  So while the speed and initial configuring setup is really nice - this seems to have broke a lot of the niceties we had from multi root and scale and the work we had done on getting configuring screens right.  I'll dig some more and try to provide crash logs and point to problem areas in the code causing the issues, but for now I just wanted to point the issues I'm seeing.

andydotxyz:
If you could provide more details that would really help.
I have tested here using a 4k laptop screen, a HiDPI monitor and a standard DPI screen.
Changing primary screens and altering user configured scale all seemed to work fine for me.
The crash info would be helpful - I have not seen any crashes and I use this branch constantly.

Can you please also provide more information about "broke a lot of the niceties we had from multi root" - what is broken?
Scale and window dragging etc all work perfectly here...

andydotxyz:
@Jacalz you mentioned testing this branch before, but did not mention any issues - can you replicate any of what @okratitan reported?

Jacalz:
@andydotxyz I did not see those issues, but both my screens are 1080p and I haven't messed about with `arandr` much.

andydotxyz:
@okratitan it's been a week since your last comment - are there any logs or reproduction steps that you can provide to show exactly the issues you are facing?
I'm kindof stuck at this point, the main changes in this PR are now 3 weeks old and I don't feel any closer to getting it landed.

okratitan (approved):

Jacalz (approved):
Works much better than before for me. Approving :)

PR #147, opened 2020-11-02
Filter background image selction and default to pictures dir

Jacalz:
This commit changes the starting location of the file dialog that selects the background to try to use the pictures folder.
It also filters out files to only show jpg/jpeg, png and svg.

image

### DIFF
--- a/internal/ui/settings_ui.go
+++ b/internal/ui/settings_ui.go
@@ -1,7 +1,6 @@
 package ui
 
 import (
-	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
@@ -378,25 +377,22 @@ func getPicturesDir() (fyne.ListableURI, error) {
 		return nil, err
 	}
 
-	if runtime.GOOS == "darwin" {
-		uri, err := storage.Child(storage.NewFileURI(home), "Pictures")
-		if err != nil {
-			return nil, err
-		}
+	const xdg = "xdg-user-dir"
+	if _, err := exec.LookPath(xdg); err == nil {
+		cmd := exec.Command(xdg, "PICTURES")
 
-		return storage.ListerForURI(uri)
+		loc, err := cmd.Output()
+		if err == nil {
+			loc = loc[:len(loc)-1] // Remove \n at the end
+			uri := storage.NewFileURI(string(loc))
+			return storage.ListerForURI(uri)
+		}
 	}
 
-	const xdguserdir, pic = "xdg-user-dir", "PICTURES"
-	cmd := exec.Command(xdguserdir, pic)
-	stdout := bytes.NewBufferString("")
-	cmd.Stdout = stdout
-	if err = cmd.Run(); err != nil {
+	uri, err := storage.Child(storage.NewFileURI(home), "Pictures")
+	if err != nil {
 		return nil, err
 	}
 
-	loc := stdout.String()
-	loc = loc[:len(loc)-1] // Remove \n at the end of string
-
-	return storage.ListerForURI(storage.NewFileURI(loc))
+	return storage.ListerForURI(uri)
 }
--- a/internal/ui/settings_ui.go
+++ b/internal/ui/settings_ui.go
@@ -381,10 +381,10 @@ func getPicturesDir() (fyne.ListableURI, error) {
 	if _, err := exec.LookPath(xdg); err == nil {
 		cmd := exec.Command(xdg, "PICTURES")
 
-		loc, err := cmd.Output()
-		if err == nil {
-			loc = loc[:len(loc)-1] // Remove \n at the end
-			uri := storage.NewFileURI(string(loc))
+		out, err := cmd.Output()
+		location := string(out[:len(out)-1]) // Remove \n at the end
+		if err == nil && location != home {
+			uri := storage.NewFileURI(location)
 			return storage.ListerForURI(uri)
 		}
 	}
--- a/internal/ui/settings_ui.go
+++ b/internal/ui/settings_ui.go
@@ -1,10 +1,10 @@
 package ui
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"runtime"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


internal/ui/settings_ui.go
```diff
@@ -60,22 +63,29 @@ func (d *settingsUI) loadAppearanceScreen() fyne.CanvasObject {
...
+
+		// not advisable for cross-platform but we are desktop only
+		path := file.URI().String()[7:]
+		// TODO add a nice preview :)
+		_ = file.Close()
+
+		bgPath.SetText(path)
+		bgPathClear.Enable()
+	}, d.win)
+	bgDialog.SetFilter(&storage.MimeTypeFileFilter{MimeTypes: []string{"image/jpg", "image/jpeg", "image/png", "image/svg"}})
```
andydotxyz, line 79: I think that the correct mime for SVG is "image/svg+xml".
However I would use the file extension filter personally - that is what you want to filter right?
If for some reason a system is missing the mime lookups this will not do what you intend.

internal/ui/settings_ui.go
```diff
@@ -361,3 +371,13 @@ func modifierToString(mods deskDriver.Modifier) string {
 	}
 	return strings.Join(s, "+")
 }
+
+func getPicturesDir() (fyne.ListableURI, error) {
+	home, err := os.UserHomeDir()
```
andydotxyz, line 376: To do this correctly will be more work I'm afraid.
The Pictures directory is part of the same XDG spec that is used to find the correct locations for the favourites folders...

`XDG_PICTURES_DIR="$HOME/Pictures"`

I don't think that code was exported from the Fyne `dialog` package in any way, so you may need to re-do the `xdg-user-dirs` lookup yourself here.

andydotxyz (changes requested):

internal/ui/settings_ui.go
Jacalz: Good point. Will switch to extensions :)

internal/ui/settings_ui.go
Jacalz: Good point. Fixed now :)

internal/ui/settings_ui.go
```diff
@@ -378,6 +378,25 @@ func getPicturesDir() (fyne.ListableURI, error) {
 		return nil, err
 	}
 
-	uri := storage.NewFileURI(filepath.Join(home, "Pictures"))
-	return storage.ListerForURI(uri)
+	if runtime.GOOS == "darwin" {
```
andydotxyz, line 381: Here we set up a darwin branch for the generic solution, then assume xdg for any other system... So if someone were running this on Windows, or Linux without the xdg binary (totally possible) it will fail with an error.

Why not invert the logic - if the exec.LookPath finds xdg-user-dir then use it, otherwise take the generic approach?

internal/ui/settings_ui.go
```diff
@@ -378,6 +378,25 @@ func getPicturesDir() (fyne.ListableURI, error) {
...
+		if err != nil {
+			return nil, err
+		}
+
+		return storage.ListerForURI(uri)
+	}
+
+	const xdguserdir, pic = "xdg-user-dir", "PICTURES"
+	cmd := exec.Command(xdguserdir, pic)
+	stdout := bytes.NewBufferString("")
```
andydotxyz, line 392: You can simplify this by using `Command.Output()` I think

andydotxyz: I realise now you would have got this from the other XDG code in Fyne - that probably needs tidied up further!

internal/ui/settings_ui.go
```diff
@@ -378,6 +378,25 @@ func getPicturesDir() (fyne.ListableURI, error) {
...
+	cmd.Stdout = stdout
+	if err = cmd.Run(); err != nil {
+		return nil, err
```
andydotxyz, line 395: This should fall back to a sensible default as well

internal/ui/settings_ui.go
```diff
@@ -378,6 +378,25 @@ func getPicturesDir() (fyne.ListableURI, error) {
...
+	}
+
+	loc := stdout.String()
+	loc = loc[:len(loc)-1] // Remove \n at the end of string
+
+	return storage.ListerForURI(storage.NewFileURI(loc))
```
andydotxyz, line 401: Note lastly that xdg-user-dir will fall back to the home directory, so should check to see if that is the case and fall back to a better default

andydotxyz (changes requested):

internal/ui/settings_ui.go
Jacalz: Sorry. I did not think that fynedesk was runnable on Windows due to relying on Xorg among other stuff.
That sounds like a great approach :+1:

internal/ui/settings_ui.go
Jacalz: I noticed that you did so. Thank you :+1:

internal/ui/settings_ui.go
Jacalz: Good point.

internal/ui/settings_ui.go
Jacalz: Good point.

internal/ui/settings_ui.go
```diff
@@ -361,3 +370,29 @@ func modifierToString(mods deskDriver.Modifier) string {
...
+	if err != nil {
+		return nil, err
+	}
+
+	const xdg = "xdg-user-dir"
+	if _, err := exec.LookPath(xdg); err == nil {
+		cmd := exec.Command(xdg, "PICTURES")
+
+		loc, err := cmd.Output()
+		if err == nil {
```
andydotxyz, line 385: I think this previous comment is still needing addressed:
 
> Note lastly that xdg-user-dir will fall back to the home directory, so should check to see if that is the case and fall back to a better default

You'll need to check whether or not the returned location is the home directory (like https://github.com/fyne-io/fyne/blob/0ca05dc12c03b8873d2f22caa6ce4c28988b3886/dialog/file_xdg.go#L36).

andydotxyz (changes requested):

internal/ui/settings_ui.go
Jacalz: Good point. Changed it so it will try the fallback if that is the case.

andydotxyz (approved):

PR #149, opened 2020-12-09
Add a way of running in embedded X window using make

Jacalz:
I am tired of copy-and-pasting the lines from the README whenever I want to do this. Now we can just run `make run -j2` to do it :)


andydotxyz (approved):

Jacalz:
This seems to be blocked by `staticcheck` not completing and thus needs https://github.com/fyne-io/fynedesk/pull/144 to land first or this could alternatively land manually.

PR #150, opened 2021-01-03
Make staticcheck happy for the fyne 2.0 update

Jacalz:
Fixes a bunch of deprecated function usages.

### DIFF
--- a/internal/ui/settings_ui.go
+++ b/internal/ui/settings_ui.go
@@ -93,9 +93,9 @@ func (d *settingsUI) loadAppearanceScreen() fyne.CanvasObject {
 	clockFormat.SetSelected(d.settings.ClockFormatting())
 
 	themeLabel := widget.NewLabel(d.settings.IconTheme())
-	themeIcons := fyne.NewContainerWithLayout(layout.NewHBoxLayout())
+	themeIcons := container.NewHBox()
 	d.populateThemeIcons(themeIcons, d.settings.IconTheme())
-	themeList := fyne.NewContainerWithLayout(layout.NewVBoxLayout())
+	themeList := container.NewVBox()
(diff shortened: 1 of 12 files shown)


internal/ui/settings_ui.go
```diff
@@ -105,25 +105,25 @@ func (d *settingsUI) loadAppearanceScreen() fyne.CanvasObject {
...
+	time := container.NewBorder(nil, nil, clockLabel, clockFormat)
 	top := container.NewVBox(bg, time)
 
 	themeFormLabel := widget.NewLabelWithStyle("Icon Theme", fyne.TextAlignCenter, fyne.TextStyle{Bold: true})
-	themeCurrent := widget.NewHBox(layout.NewSpacer(), themeLabel, themeIcons)
-	bottom := fyne.NewContainerWithLayout(layout.NewBorderLayout(nil, themeCurrent, themeFormLabel, nil), themeCurrent, themeFormLabel, container.NewScroll(themeList))
+	themeCurrent := container.NewHBox(layout.NewSpacer(), themeLabel, themeIcons)
+	bottom := container.NewBorder(nil, themeCurrent, themeFormLabel, nil, container.NewScroll(themeList))
 
-	applyButton := widget.NewHBox(layout.NewSpacer(),
```
andydotxyz, line 116: Doesn't removing this spacer break the right alignment?

andydotxyz (changes requested):
Thanks so much for doing this.
Just one removed spacer that I think was a mistake.

internal/ui/settings_ui.go
Jacalz: Good catch. That was not intentionally. Will get it sorted tomorrow.

andydotxyz (approved):

PR #154, opened 2021-01-30
Allow resize from top corners

### LINKED ISSUE #113: Allow to resize window that overflows
sometimes windows are too big for the screen (an other issue could be open for that)
AFAIK there is only one way to reduce height of window, and if the bottom bar is out of the screen because the app overflows, then we can't do it anymore.

We could add a shortcut such as alt-click or alt-F7 to move window over the top to be able to grab the bottom or we could allow to grab the top border to resize.

Both solution should be implemented from my point of view

andydotxyz:
Just a narrow strip, avoiding buttons.
FIxes #113


andydotxyz:
No comments in defaulting time, auto-accepted

PR #155, opened 2021-01-30
Render menu over windows

### LINKED ISSUE #128: Pop account menu over windows
Use the new splash window handling to appear above windows...

andydotxyz:
Appears and works, but does not dismiss unless clicked.

Possible solutions one of the following:
- [x] Dismiss when mouse leaves menu
- [ ] DIsmiss if user clickes outside menu

Fixes #128


okratitan (approved):
Solid and how I would have done it too. LGTM

PR #156, opened 2021-02-08
Fix dnd with chrome

### LINKED ISSUE #49: DND in Chrome not dropping or hovering
This doesn't work at the moment - a new window for the tab you start dragging is created.

okratitan:
This fixes dnd with chrome and fixes #49

### DIFF
--- a/wm.go
+++ b/wm.go
@@ -14,13 +14,13 @@ type WindowManager interface {
 }
 
 // Stack describes an ordered list of windows.
-// The order of the windows in this list is bottom to top from what is visible on screen.
-// TopWindow() returns the last element in the stack which is the top most window on screen.
+// The order of the windows in this list matches the stacking order on screen.
+// TopWindow() returns the 0th element with each item after that being stacked below the previous.
 type Stack interface {
 	AddWindow(Window)    // Add a new window to the stack
-	RaiseToTop(Window)   // Request that the passed window become top of the stack
+	RaiseToTop(Window)   // Request that the passed window become top of the stack.
 	RemoveWindow(Window) // Remove a specified window from the stack
-	TopWindow() Window   // Get the top most window currently visible
+	TopWindow() Window   // Get the currently top most window
 	Windows() []Window   // Return a list of all managed windows. This should not be modified
 }
 
--- a/wm.go
+++ b/wm.go
@@ -14,13 +14,13 @@ type WindowManager interface {
 }
 
 // Stack describes an ordered list of windows.
-// The order of the windows in this list matches the stacking order on screen.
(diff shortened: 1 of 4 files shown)


okratitan:
Before approving we should discuss if reversing the stack list is correct - or if we should use the bottom to top order in stack list to begin with.  I didn't make this change yet because I couldn't remember if there a reason we went to top to bottom on stack list.

andydotxyz:
Great catch. As for the ordering I am not sure. I think the window order was based on the order we would display them on switcher. So either we reverse it here, or in the switcher code. We should probably document the order in the `Stack` interface I guess.

andydotxyz (dismissed):

wm.go
```diff
@@ -14,13 +14,13 @@ type WindowManager interface {
 }
 
 // Stack describes an ordered list of windows.
-// The order of the windows in this list matches the stacking order on screen.
-// TopWindow() returns the 0th element with each item after that being stacked below the previous.
```
andydotxyz, line 18: Wow, I forgot that we defined it this way around.
Are we certain that reversing it for all platforms is the right thing to do?

andydotxyz:
I jumped over to WinGo as an example of another WM in go - they have stacked it how we previously did:

https://github.com/BurntSushi/wingo/blob/30b336cbb88d32c12c14768d5ff89b3d55560081/stack/stack.go

okratitan:
Yes and they had to provide a helper function to convert it.  Walkinh the slice to convert the stack changes on every focus, raise, lower, add, and remove of windows surely isn't ideal, and appending new windows or raised windows to the end of a slice surely makes more sense than inserting them at the beginning.

wm.go
okratitan: I can't see any reason why it wouldn't be fine for any other platform -- I'm not certain though.  However I'm less certain that having it top to bottom is the right thing to do.  So far we have one implementation - X and it says bottom to top.  That should be fine for Wayland from what I can tell so far too...  I would say between the two there isn't a lot of certainty of which one is better - but the one concrete thing we have and know is how the one platform we have wants it bottom to top and from a code standup that means less conversions and slice walking so I think that tips that scales for me.

andydotxyz:
If The X11 driver code wants to use the reverse order for performance reasons then that is fair enough.
But that in itself does not indicate that we should reverse the exposed API...

Another approach would be that the `Windows` method could reverse it so that the backwards list is internal to the driver only and our API remains unaltered.

okratitan:
Fair enough - reverted the interface/switcher changes and reverse the stack on Windows() method

andydotxyz (approved):
Thanks, this feels like a nice clean change and a very welcome fix :)

PR #164, opened 2021-02-12
A few improvements to the screenshot shower

Jacalz:
Below is a list with the changes:
- A "screenshot.png" filename is now preset when saving.
- The default folder for saving is now the Pictures dir (if found).
- The file is now propertly closed after being saved.
- Various code cleanups for readability.

### DIFF
--- a/internal/ui/screenshot.go
+++ b/internal/ui/screenshot.go
@@ -3,6 +3,7 @@ package ui
 import (
 	"image"
 	"image/png"
+	"time"
 
 	"fyne.io/fyne/v2"
 	"fyne.io/fyne/v2/canvas"
@@ -75,7 +76,9 @@ func saveImage(pix image.Image, w fyne.Window) {
 	}, w)
 
 	d.SetFilter(storage.NewMimeTypeFileFilter([]string{"image/png"}))
-	d.SetFileName("screenshot.png")
+
+	now := time.Now().Format("20060102T150405") // YYYYMMDD"T"HHMMSS
+	d.SetFileName("screenshot-" + now + ".png")
 
 	if dir, err := getPicturesDir(); err == nil {
 		d.SetLocation(dir)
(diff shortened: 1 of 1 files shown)


okratitan (changes requested):
I think screenshot + time.now() would be even better.  That way you can just screenshot and never worry about naming the file or overwriting if you just want to click save and the file names would then be easy to sort by time too.

AlbinoGeek:
> I think screenshot + time.now() would be even better. That way you can just screenshot and never worry about naming the file or overwriting if you just want to click save and the file names would then be easy to sort by time too.

I would prefer a much more standardized date format (for humans) than a unix timestamp, such as:

> The representation of ISO date formats does away with the need for country-specific notations, at least in electronic formats. This way, time and date information is standardized, resulting in fewer communication issues.  The ISO 8601 standard is also known as ANSI INCITS 30-1997 (R2008) or NIST FIPS PUB 4-2 in the United States.

> According to the ISO standard, all values like times, dates or durations are shown in a certain order: They start with the largest unit and proceed to the next smaller value. This sequence is known as “descending notation”. It corresponds with the natural mathematic value of numbers. Larger units are therefore always written before smaller units. This is advantageous as it means the lexicographical and chronological sorting of dates and times will produce the same result.

Here is an example using ISO 8601 "Basic Format" for use in filenames:

```go
  // Safest (in regards to filename-safe characters) and simplest, chronological sorting with A-Z, etc.
  time.Now().Format("20060102T150405") // YYYYMMDD"T"HHMMSS
  
  // With milliseconds (not necessary for humans):
  time.Now().Format("20060102T150405.000") // YYYYMMDD"T"HHMMss.fff

  // With a specified timezone:
  time.Now().Format("20060102T150405.000-0700") // YYYYMMDD"T"HHMMss.fff-hhmm
```

These formats are **not compatible with RFC 3339** and that is okay. Just don't confuse the two.

And if you wanted to go "full human", but losing the lexical benefits (like sorting):

```go
  // If using a date format that uses words, consider localization using `monday`:
  monday.Format(time.Now(), "Jan 02 2006 3:04:05 PM", monday.LocaleEnUS)
```

okratitan (approved):
LGTM

Issue #119, opened 2020-06-30
Windows are leaving trails behind


Jacalz:
Moving windows around causes trails to be left behind. Windows also feel "floaty" and do not follow the mouse when moving the mouse around quickly. Please look at the archive below for a video of what happens:

[trailing-and-floating-windows.zip](https://github.com/fyne-io/fynedesk/files/4852129/trailing-and-floating-windows.zip)

andydotxyz:
Double check with latest develop as it's on FYne v1.3.2 now :)

Jacalz:
It looks like it didn’t improve much sadly. Windows are still trailing a lot :(

okratitan:
This should be resolved by two upcoming fixes we are aware of - the vsync stuff and concurrent configurewindow changes

okratitan:
Leaving this open - I think we solved the feel "floaty" and don not follow the mouse when moving quickly portion of this but there still seems to be trails for @Jacalz when he is moving windows on the primary screen with the fyne window underneath.

andydotxyz:
Pretty sure that adding compton will have fixed these. If you install that binary it will be a proper compositor.
In the future we will integrate similar code so it does not need an external app.

PR #165, opened 2021-02-13
Use a queue to configure windows on move/resize.

### LINKED ISSUE #163: Applications can push other ones off the screen
Moving around windows sometimes allows for pushing other ones off the screen. Video was too big to upload here, please see https://gophers.slack.com/archives/CKV7AF2N6/p1613061709261800.

okratitan:
Don't propogate mouse clicks on moveresize events
Don't send a configure after moveresize cancel

Fixes #163 and #119

### DIFF
--- a/internal/x11/win/client.go
+++ b/internal/x11/win/client.go
@@ -185,7 +185,7 @@ func (c *client) NotifyBorderChange() {
 }
 
 func (c *client) NotifyGeometry(x int, y int, width uint, height uint) {
-	c.frame.updateGeometry(int16(x), int16(y), uint16(width), uint16(height), true)
+	c.frame.queueGeometry(int16(x), int16(y), uint16(width), uint16(height), true)
 }
 
 func (c *client) NotifyFullscreen() {
@@ -224,6 +224,7 @@ func (c *client) NotifyMouseRelease(x, y int16, b xproto.Button) {
 }
 
 func (c *client) NotifyMoveResizeEnded() {
+	c.frame.closeQueueGeometry()
 	c.frame.notifyInnerGeometry()
 }
 
--- a/internal/x11/win/client.go
+++ b/internal/x11/win/client.go
@@ -184,11 +184,7 @@ func (c *client) NotifyBorderChange() {
 	}
 }
 
-func (c *client) NotifyGeometry(x int, y int, width uint, height uint, queue bool) {
-	if queue {
-		c.frame.queueGeometry(int16(x), int16(y), uint16(width), uint16(height), true)
-		return
-	}
+func (c *client) NotifyGeometry(x int, y int, width uint, height uint) {
 	c.frame.updateGeometry(int16(x), int16(y), uint16(width), uint16(height), true)
 }
 
@@ -228,7 +224,7 @@ func (c *client) NotifyMouseRelease(x, y int16, b xproto.Button) {
 }
 
 func (c *client) NotifyMoveResizeEnded() {
-	c.frame.closeQueueGeometry()
+	c.frame.endConfigureLoop()
 	c.frame.notifyInnerGeometry()
 }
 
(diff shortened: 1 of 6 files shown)


andydotxyz:
I think ths shutting down of the configure loop is a little unsafe, consider this alternative approach:

```go
diff --git a/internal/x11/win/frame.go b/internal/x11/win/frame.go
index add38ba..26ee060 100644
--- a/internal/x11/win/frame.go
+++ b/internal/x11/win/frame.go
@@ -233,6 +233,11 @@ func (f *frame) checkScale() {
 func (f *frame) configureLoop() {
        var lastGeometry configureGeometry
        var change = false
+       defer func() {
+               close(f.configureChan)
+               f.configureChan = nil
+               f.configureQuitChan = nil
+       }()
        for {
                select {
                case g := <-f.configureChan:
@@ -252,11 +257,8 @@ func (f *frame) configureLoop() {
 }
 
 func (f *frame) closeQueueGeometry() {
-       if f.configureChan != nil {
-               close(f.configureChan)
+       if f.configureQuitChan != nil {
                f.configureQuitChan <- struct{}{}
-               f.configureChan = nil
-               f.configureQuitChan = nil
        }
 }
```

andydotxyz:
Or an even simpler version that does not use the quit channel:

```go
diff --git a/internal/x11/win/frame.go b/internal/x11/win/frame.go
index add38ba..b13ba2d 100644
--- a/internal/x11/win/frame.go
+++ b/internal/x11/win/frame.go
@@ -44,7 +44,6 @@ type frame struct {
        cancelFunc context.CancelFunc
 
        configureChan     chan configureGeometry
-       configureQuitChan chan struct{}
 
        canvas test.WindowlessCanvas
        client *client
@@ -236,10 +235,12 @@ func (f *frame) configureLoop() {
        for {
                select {
                case g := <-f.configureChan:
+                       if g.x == 0 && g.y == 0 && g.width == 0 && g.height == 0 { // closing
+                               f.configureChan = nil
+                               return
+                       }
                        lastGeometry = g
                        change = true
-               case <-f.configureQuitChan:
-                       return
                default:
                        if change && f.configureChan != nil {
                                if lastGeometry.x != 0 && lastGeometry.y != 0 && lastGeometry.width != 0 && lastGeometry.height != 0 {
@@ -254,9 +255,6 @@ func (f *frame) configureLoop() {
 func (f *frame) closeQueueGeometry() {
        if f.configureChan != nil {
                close(f.configureChan)
-               f.configureQuitChan <- struct{}{}
-               f.configureChan = nil
-               f.configureQuitChan = nil
        }
 }
 
@@ -803,7 +801,6 @@ func (f *frame) unmaximizeApply() {
 func (f *frame) queueGeometry(x int16, y int16, width uint16, height uint16, force bool) {
        if f.configureChan == nil {
                f.configureChan = make(chan configureGeometry, 5)
-               f.configureQuitChan = make(chan struct{})
                go f.configureLoop()
        }
        f.configureChan <- configureGeometry{x, y, width, height, force}
```

andydotxyz (changes requested):
Naming is a little inconsistent here - i.e. `configureLoop` is closed with `closeQueueGeometry`.

Also the `true` parameter to `NotifyGeometry` meaning "queue" seems unintuitive, indicating that maybe we were over-using the notification method. Perhaps adding `RequestGeometry` or `QueueGeometry` would be clearer in the code?

One last comment - could there be a clearer name than 'configureChan' that doesn't repeat the type of the variable? Maybe `pendingGeometry`?

Jacalz (commented):
Thanks for this! Things are much much smoother, but the windows are still leaving trails behind unfortunately. The slowness when dragging windows is entirely gone though.

https://user-images.githubusercontent.com/25466657/107883493-9ae6fb00-6eef-11eb-8c7d-dd0b74658efa.mp4

Jacalz (approved):
I am approving this as it is working great and the code looks good to me. It does, like said before, not fix the trailing windows, but things are way more responsive. I think you still might want @andydotxyz to review before it lands as he has much better understanding of the code.

andydotxyz (approved):
Nice speed increase thanks

PR #166, opened 2021-02-13
Move key bindings from Alt to Super and fix issue when they could stop working

### LINKED ISSUE #160: alt-space not reliable
After some times the alt-space shortcut stop working, other shortcuts work fine so I guess it's specific to the application window.

Tell me if I can help debugging

andydotxyz:
Problem was visible when all windows were closed - was a focus issue.

Fixes #160

### DIFF
--- a/internal/ui/desk.go
+++ b/internal/ui/desk.go
@@ -223,13 +223,13 @@ func (l *desktop) addSettingsChangeListener() {
 }
 
 func (l *desktop) registerShortcuts() {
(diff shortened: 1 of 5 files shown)


okratitan (approved):

Jacalz (approved):
Thanks. Looks good to me. We might want to have a more fleshed out application launcher and only use "super" to open it, but that's certainly future work.

PR #167, opened 2021-02-14
Add status commands to launch window

andydotxyz:
Add volume controls.
Fixed some volume bugs as well.


okratitan (approved):
really neat idea thanks!

PR #168, opened 2021-02-14
Make modifier Alt/Super a user preference

andydotxyz:
Just what it says, allow users to choose Alt for their standard key bindings instead of Super.


okratitan (approved):
Looks good

PR #170, opened 2021-02-17
Half working left or right toolbar button pos

Jacalz:
This is unfortunately as far as I could get on my own. There are various issues that I have been unable to solve.

- The windows do not resize correctly when using right position.
- Already open windows do not update when applying the setting in settings.

- The correct selected values are not shown in settings on start. Seems to be an issue in the clock as well so not related to this PR, I guess.

### DIFF
--- a/wm/border.go
+++ b/wm/border.go
@@ -53,20 +53,42 @@ func NewBorder(win fynedesk.Window, icon fyne.Resource, canMaximize bool) *Borde
 	}}
 
 	title := canvas.NewText(win.Properties().Title(), theme.ForegroundColor())
-	titleBar := newColoredHBox(win.Focused(), win, makeFiller(0),
-		newCloseButton(win),
-		max,
-		min,
-		title,
-		layout.NewSpacer())
+	buttonPos := fyne.CurrentApp().Preferences().StringWithFallback("toolbarbuttonposition", "Left")
+
+	var titleBar *Border
+	if buttonPos == "Left" {
+		titleBar = newColoredHBox(win.Focused(), win, makeFiller(0),
+			newCloseButton(win),
+			max,
+			min,
+			layout.NewSpacer(),
+			title,
+			layout.NewSpacer(),
+		)
+	} else {
+		titleBar = newColoredHBox(win.Focused(), win, makeFiller(0),
+			layout.NewSpacer(),
+			title,
+			layout.NewSpacer(),
+			min,
+			max,
+			newCloseButton(win),
+		)
+	}
+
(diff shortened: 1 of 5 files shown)


wm/border.go
```diff
@@ -53,20 +53,42 @@ func NewBorder(win fynedesk.Window, icon fyne.Resource, canMaximize bool) *Borde
 	}}
 
 	title := canvas.NewText(win.Properties().Title(), theme.ForegroundColor())
-	titleBar := newColoredHBox(win.Focused(), win, makeFiller(0),
-		newCloseButton(win),
-		max,
-		min,
-		title,
-		layout.NewSpacer())
+	buttonPos := fyne.CurrentApp().Preferences().StringWithFallback("toolbarbuttonposition", "Left")
```
Jacalz, line 56: I could not really figure out how to get this working using the settings code. Any advice would be appreciated.

andydotxyz:
I guess this should be a draft PR for now?
I'll see if I can look into the settings code.

wm/border.go
andydotxyz: This looks OK - what part of it isn't working as you'd like?

andydotxyz:
I can't see anything wrong, can you expand on the problem you were encountering @Jacalz ?

wm/border.go
Jacalz: I just thought that I'd be able to use the actually DeskSettings struct, but if this is the good way to go, I'm happy with that.

Jacalz:
> I can't see anything wrong, can you expand on the problem you were encountering @Jacalz ?

Sure thing. The resizing of windows is entirely broken when using window decorations on the right side. I've tried to put the spacers in different places but to no avail. Having them on the left works as expected.  

https://user-images.githubusercontent.com/25466657/109161196-0f832a80-7777-11eb-9a1a-f600eec60769.mp4

There is of course also the issue with windows not updating when the left/right setting change. You have to open a new window for it to show.

andydotxyz:
Ah yes I see. There is an optimisation in the code that chops the image in two so it can be resized without redrawing.
If you move the chop position in appropriately it should be fixed.

andydotxyz:
See `internal/x11/win` around 345:

`iconAndBorderSizePix := iconSizePix + x11.BorderWidth(x11.XWin(f.client))*2`

andydotxyz:
This should now be completed on `feature/decorations` branch for perusal before pulling in here/opening new PR.

Jacalz:
Closing this in favor of a new PR coming up shortly...

PR #172, opened 2021-02-17
Add the ability to unmaximize a window by dragging.

### LINKED ISSUE #138: Window not draggable out of maximized state
It seems to currently not be possible of dragging a window out of the maximized state. If I maximized the window (either by double clicking the title bar or pressing the button) I seem to need to do the same action in reverse to stop the window from being maximized.

okratitan:
Fixes #138


internal/x11/win/frame.go
```diff
@@ -484,16 +485,31 @@ func (f *frame) maximizeApply() {
...
 	}
 	moveDeltaX := x - f.mouseX
 	moveDeltaY := y - f.mouseY
-	f.mouseX = x
-	f.mouseY = y
 	if moveDeltaX == 0 && moveDeltaY == 0 {
 		return
 	}
+	if f.client.Maximized() {
+		if uint16(moveDeltaX) > x11.TitleHeight(x11.XWin(f.client)) ||
```
andydotxyz, line 497: Might it be clearer if we add a new variable that avoids this call twice? Might help show intent too.
```go
dragThreshold := x11.TitleHeight(x11.XWin(f.client))
```

andydotxyz (changes requested):
This is a great change thanks - just a suggestion inline let me know what you think.

internal/x11/win/frame.go
okratitan: With the new changes it's only one threshold now, vertical.  Per @Jacalz horizontal drag shouldn't unmaximize.  That will change we do quarter snapping though I'm sure.

andydotxyz: There is now a magic "*3" so I think the comment still applies with regard to intent.
Looking at the code it is also not clear why the y position needs to be checked twice.

Jacalz (approved):
Feels great. Good job :+1:

internal/x11/win/frame.go
andydotxyz: Thanks for the explanation, if the first check is to ensure it is within the bar then you’re missing a check for y >= f.y.

I don’t think a const makes sense, as the title height is a variable.
However the same border theme lookup is being done twice in this line, so is it not more efficient to capture in a var?

andydotxyz (approved):
Great improvement thanks

PR #176, opened 2021-02-24
Add AC power indicator to battery module

andydotxyz:
Just what it says on the tin :) Updates along with the value - electric bold if it's plugged in, otherwise shows the battery

screenshot-20210224T165225


andydotxyz:
Ugh, no idea why the fyne tool is generating code that fails staticcheck here :(

modules/status/battery_other.go
```diff
@@ -11,6 +11,15 @@ import (
 	"fyne.io/fyne/v2"
 )
 
+func (b *battery) powered() (bool, error) {
+	status, err := ioutil.ReadFile("/sys/class/power_supply/BAT0/status")
```
okratitan, line 15: Is there potential that there are multiple batteries and BAT1 is currently the one working?

okratitan (commented):
LGTM just one question

modules/status/battery_other.go
andydotxyz: This is true, I did not know how to read the AC status outside battery info. Do you know?

okratitan: so status should be "Charging" if a battery is charging.  So perhaps something like this pseudocode:
charging := false
subdirs := ls("/sys/class/power_supply")
for _, subdir := range subdirs {
        status := read("/sys/class/power_supply"+subdir+"/status")
        if status == "Charging" {
                charging = true
                break
        }
}
if charging { b.icon = charging } else { b.icon = not charging }

okratitan: this would check all batteries and if none are charging then set the battery icon to not charging... if even one is charging then set the battery icon to charging.  That's how you would have to do it using sysfs... Udev and Upower provide better information related to this than sysfs but I don't think you are wanting to implement a new subsystem at this point.

okratitan: Side note -- I haven't looked at how you are calculating battery remaining or percentages yet either... but you potentially need to be doing the same thing for that... adding the values for each battery

okratitan (approved):

modules/status/battery_other.go
okratitan: due to the design being one battery currently we will land this change as is assuming one battery and explore updating the entire module to no longer make that assumption in the future.

PR #178, opened 2021-03-08
New "Swedish Pine" background for next release

Jacalz:
This replaces the old "Loch Fyne" background with a new "Swedish Pine" background.
The image is in 4K resolution and is roughly half the size of the old image (uses 80% jpeg quality and no chroma subsampling).

I also removed an unused linter directive (for staticcheck) that should have been removed a while ago.


andydotxyz (approved):
Looks great on my HiDPI

PR #179, opened 2021-03-09
This backports CI changes from fyne 

Jacalz:
The setup of each workflow run should now be about 3x faster thanks to using https://github.com/WillAbides/setup-go-faster. We should also be getting inline staticcheck recommendations for PRs. This includes a workaround until WillAbides/setup-go-faster#11 is fixed.

This also makes sure that we run tests on the latest stable Go release (Go 1.16). It has the side-effect that the CI should be even faster: "Linking is 20-25% faster than 1.15 and requires 5-15% less memory on average for linux/amd64, with larger improvements for other architectures and OSes.".

I tried to get golangci-lint working, but it just kept failing for weird resons. I see the same errors being partially viable in fyne, so we might have to update that to stop using golangci-lint. Annotations in https://github.com/fyne-io/fyne/actions/runs/633663817 are more or less the same errors that made golangci-lint useless here.


Jacalz:
It was not my intention to open the branch in this repo and not my fork. I did the change through the web as I find it to generally be easier for GitHub actions changes, and GitHub apparently opened the branch here. Sorry...

andydotxyz (approved):

PR #188, opened 2021-05-28
Allow fyneterm to be picked up automatically

Jacalz:
This lets the default pinned terminal pick up fyneterm as an option if it is installed on the system.
If it is not, it will fall back to the other common terminals that might exist.

### DIFF
--- a/internal/icon/fdo.go
+++ b/internal/icon/fdo.go
@@ -542,7 +542,7 @@ func appendAppIfExists(apps []fynedesk.AppData, app fynedesk.AppData) []fynedesk
 func (f *fdoIconProvider) DefaultApps() []fynedesk.AppData {
 	var apps []fynedesk.AppData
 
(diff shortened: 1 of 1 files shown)


andydotxyz:
> I could not quite decide on if we wanted it first or last in the list. Added it to the beginning, but I can change the position if you feel it makes sense.

Quite right ;)

andydotxyz (approved):
Nice improvement, but I think you’ll need to merge in the CI fixes before this will merge

Jacalz:
Looks like the merge brought in some stray old commit which shouldn’t be there I think. Will have to sort this out tomorrow.

andydotxyz:
Looks like this was a master merge in instead of develop too.

Jacalz:
Fixed now. Sorry for the inconvenience.

andydotxyz (approved):
Thanks

PR #193, opened 2021-06-09
Fix the bug of main menu display position when Runner is not used

myml:
Main menu can not show in embed mode. Position error due to height error.

### DIFF
--- a/internal/x11/wm/desk.go
+++ b/internal/x11/wm/desk.go
@@ -655,8 +655,12 @@ func (x *x11WM) showWindow(win xproto.Window, parent xproto.Window) {
 		xproto.ChangeWindowAttributes(x.Conn(), win, xproto.CwEventMask, []uint32{xproto.EventMaskLeaveWindow})
 
 		screen := fynedesk.Instance().Screens().Primary()
-		//attrs, _ := xproto.GetGeometry(x.Conn(), xproto.Drawable(win)).Reply()
-		w, h := 200*screen.CanvasScale(), 169*screen.CanvasScale() // TODO figure out why this lookup does not work for BSD: int(attrs.Width), int(attrs.Height)
+		// attrs, _ := xproto.GetGeometry(x.Conn(), xproto.Drawable(win)).Reply()
+		var height float32 = 136
+		if usingRunner() {
+			height = 169
+		}
+		w, h := 200*screen.CanvasScale(), height*screen.CanvasScale() // TODO figure out why this lookup does not work for BSD: int(attrs.Width), int(attrs.Height)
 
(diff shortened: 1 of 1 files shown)


andydotxyz:
I'm not sure about this. You say the error is with embedded mode, but this code only runs when it's a full desktop...

myml:
@andydotxyz  See https://github.com/fyne-io/fynedesk/blob/master/Makefile#L22

andydotxyz:
Oh I see, that embed - apologies I did not understand.
But it works OK when running as a full desktop?

myml:
> Oh I see, that embed - apologies I did not understand.
> But it works OK when running as a full desktop?

Yea! I have tested in X11 full desktop.

Befer: The main menu is located above the buttons, trigger mouse-leave event when menu is displayed, Then the menu is hidden.(The mouse must move quickly to the top after clicking)
Screenshot_deepin_2021-06-19_22:36:41
After: The main menu Normal display.
Screenshot_deepin_2021-06-19_22:37:32

andydotxyz:
Ah I see, this relates to whether or not you are using fynedesk_runner.
With runner there is an additional menu item which is why it is normally taller.

andydotxyz:
If you could update this to check:
```
	height := 124
	if os.Getenv("FYNE_DESK_RUNNER") != "" {
		height = 169
	}
```
then it should work in both instances.

myml:
> If you could update this to check:
> 
> ```
> 	height := 136
> 	if os.Getenv("FYNE_DESK_RUNNER") != "" {
> 		height = 169
> 	}
> ```
> 
> then it should work in both instances.
> (note that in my testing 136 seemed to be the minimum height rather than 124)

You're right! I did not use runner, please see the latest code.

andydotxyz (changes requested):
This looks great, but we keep functions in alphabetical order (grouped by type).
So could you please move `isRunner` up to above `max` (or rename it `usingRunner` as that is later in the alphabet ;) ).

myml:
> This looks great, but we keep functions in alphabetical order (grouped by type).
> So could you please move `isRunner` up to above `max` (or rename it `usingRunner` as that is later in the alphabet ;) ).

It's simple.

andydotxyz:
Sorry I did not notice before, it should be on the `develop` branch.
Also, for future PRs, please avoid force-pushing once a review has been added - it can make reviewing more difficult.

andydotxyz (approved):

andydotxyz:
> Sorry I did not notice before, it should be on the `develop` branch.

I was able to rebase for you this time :)

PR #195, opened 2021-08-01
Clean up the calculator module

Jacalz:
I cleaned up the code to only compile the regular expressions once instead of on each new module creation.

### DIFF
--- a/modules/launcher/calc.go
+++ b/modules/launcher/calc.go
@@ -12,14 +12,17 @@ import (
 	"github.com/Knetic/govaluate"
 )
 
+var (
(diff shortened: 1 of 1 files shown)


Jacalz:
FYI, this is also something that I plan on doing for FyneCalc.

andydotxyz:
Looks good, let's revise after we can move past 1.12 later this month.

Jacalz:
This actually seems to make the binary around 1.5 megabyte larger. I reverted the changes to use `gval` and went back to only contain the cleanups.

andydotxyz (approved):

PR #196, opened 2021-08-03
Switch from fmt.Sprintf to strconv where useful

Jacalz:
This switches to strconv where it is possible to do so without severely hurting readabilty and where it generally makes sense.
Performance and memory usage should generally be slightly better than before in these cases and there were even two cases for the sound module, where I managed to remove the convertions entirely.


andydotxyz (approved):
Nice improvements thanks

PR #199, opened 2021-08-09
Update fyne to 2.0.4

Jacalz:
This updates the fyne version to 2.0.4 from 2.0.2 which adds a lot of nice fixes and improvements :)


andydotxyz (changes requested):
I think you forgot to run the `go mod vendor` command (and watch out for that GL file getting removed!)

Jacalz:
> I think you forgot to run the `go mod vendor` command (and watch out for that GL file getting removed!)

Great catch. I think we need to update the PR template with that, just like we did in regular fyne :)

Jacalz:
I got the vendor folder updated and added in that file as well (wasn't deleted here as it didn't exist here before) :)

andydotxyz (approved):

andydotxyz:
> I got the vendor folder updated and added in that file as well (wasn't deleted here as it didn't exist here before) :)

Nice catch, thanks.

PR #200, opened 2021-08-21
Add a basic window menu

andydotxyz:
Also needed to refactor some position code that was missing...

screenshot-20210821T112730


andydotxyz:
history isn't looking right

PR #201, opened 2021-08-21
Add a basic window menu

andydotxyz:
Required tidying up some missing position handle code, all feeling a bit neater.

screenshot-20210821T112730


Jacalz:
Should it perhaps say "Minimize" instead of "Iconify"?

PR #202, opened 2021-08-21
Add a keyboard shortcut to open calculator

andydotxyz:
Maybe this should be configurable - probably in some default apps area we will add later


Jacalz (approved):
Works flawless. Nice work :)

PR #204, opened 2021-08-29
Add completed new menu

andydotxyz:
Shows the previous items as icons, a list of 5 recent apps, and all other apps categorised as metadata indicates.

Fixes: #204

menu


Jacalz (commented):
Looking really good. Just a few suggestions on improvements. The order of the categories seems to be a bit random currently. Feels like it could make more sense if it was either in alphabetical alphabetical. Recent is already at the top, which makes sense, but I think other would make sense at the bottom.

Jacalz:
Btw, seems like there is a conflict that needs to be resolved.

Jacalz (changes requested):
Sorry, two notes about this. I would still prefer to have some sort of more logical order to the categories. Something like alphabetical would be great. Would also be nice if the icons and text for applications could be left-aligned instead of centered.

andydotxyz:
Updated, looks like this now:
ordered-menu

Jacalz (approved):
Looks good to me. Nice work :+1:

PR #220, opened 2022-01-14
Position windows that are transient inside their parent

### LINKED ISSUE #48: Handle dialog type windows properly
when windowTypeGet() returns dialog we need to center the dialog on the window it is windowTransientGet() for.  Other window types may need similar attention?

andydotxyz:
This also fixes positioning of windows inside active screen when not top/left monitor

Fixes #48

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
No comments in quite some time, approving

PR #222, opened 2022-02-18
Update workflows against upstream Fyne

Jacalz:
### Description:

These are basically equivalent except for using Go 1.16 for static analysis still as I don't have 1.17 to commit the new format for build constraints.

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.


andydotxyz (approved):

Jacalz:
Strange then. What we have is at least a step up and partially working then. Let's hope that they start sending out comments in the future :)

PR #223, opened 2022-04-02
Feature: mini bar

andydotxyz:
Provide an alternative, narrow, left and right, layout for the desktop

- [x] Smaller widget panel
- [x] Adapt modules to fit
- [x] Add option for bar to match
- [x] Render notifications elsewhere so they can be read
- [ ] Pop widgets out so the full functionality can be accessed (separate PR?)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


andydotxyz:
Merging after defaulted review period

PR #225, opened 2022-05-25
Fix shortcuts with numlock

### LINKED ISSUE #224: Shortcuts are not working when numlock is on
When we have num lock activated, ev.State contains the numlock value so the shortcuts are not triggered (alt-space, alt-L etc.)

### Device (please complete the following information):
 - **OS:** Debian
 - **Version:** Sid
 - **Go version:** 1.18
 - **FyneDesk version:** develop

mattboll:
It checks for numlock value in event state and remove it if found.
It's a working solution for me but I'm not sure it's the best way to do it nor if it breaks anything

Fixes #224 

### Checklist:

- [ ] Tests included.
- [ ] Lint and formatter run with no errors.
- [ ] Tests all pass.


andydotxyz (approved):
Looks good to me I think.
The alternative is to `&` with the inverse of `ModMask2` but I think the effect is the same

PR #227, opened 2022-07-02
Show battery level in the icon

andydotxyz:
As it says on the tin :)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

100%
battery_full

25%
image


andydotxyz:
In case it's not super clear this is really helpful in the narrow-bar mode as the progress widget is not shown

Jacalz:
Would it make sense to have the battery icon  rotated 90 degrees and have the percentage shown as text within it (at least for the minibar)?

andydotxyz:
To make that work it would have to be a much larger icon, which will make it out-of-place with the others.
Or very small text.

andydotxyz:
Is that thumbs up a "will approve" or "needs work" - not sure what the current status is sorry @Jacalz

Jacalz:
Ah, sorry. I was in a hurry and just wanted to indicate that I think you were right.

andydotxyz:
Ah I see thanks.
I'll leave it here in case someone is able to approve.

andydotxyz:
Review period over

PR #234, opened 2022-12-28
Default background now matching the theme light/dark

andydotxyz:
Abstract image from Jost Grant, thanks!

### Checklist:

- [x] Tests included.
- [x] Tests all pass.


README.md
```diff
@@ -9,14 +9,14 @@
   <a href='https://coveralls.io/github/fyne-io/fynedesk?branch=develop'><img src='https://coveralls.io/repos/github/fyne-io/fynedesk/badge.svg?branch=develop' alt='Coverage Status' /></a>
 </p>
 
-# About
+# Fyne Desk
```
Jacalz, line 12: Should this not rather be something like "Fynedesk" or "Fyne Desktop"?

internal/ui/desk.go
```diff
@@ -257,10 +257,23 @@ func (l *desktop) startSettingsChangeListener(settings chan fynedesk.DeskSetting
 	}
 }
 
+func (l *desktop) startFyneSettingsChangeListener(settings chan fyne.Settings) {
+	go func() {
+		for {
+			<-settings
+			l.updateBackgrounds(l.Settings().Background())
+		}
```
Jacalz, line 265: Not a that important, but I think this would be a bit tidier if you ranged over the channel.
Something like this perhaps (I think this should work):
```go
		for range settings {
			l.updateBackgrounds(l.Settings().Background())
		}
```

internal/ui/desk.go
```diff
@@ -257,10 +257,23 @@ func (l *desktop) startSettingsChangeListener(settings chan fynedesk.DeskSetting
...
+}
+
 func (l *desktop) addSettingsChangeListener() {
 	listener := make(chan fynedesk.DeskSettings)
 	l.Settings().AddChangeListener(listener)
 	go l.startSettingsChangeListener(listener)
+
+	fyneListener := make(chan fyne.Settings)
+	l.app.Settings().AddChangeListener(fyneListener)
+	go l.startFyneSettingsChangeListener(fyneListener)
```
Jacalz, line 276: Do we need to spin up two goroutines here? The listener is in its own gorountine inside the method.

internal/x11/wm/desk.go
```diff
@@ -147,6 +148,7 @@ func NewX11WindowManager(a fyne.App) (fynedesk.WindowManager, error) {
 	go func() {
 		for {
 			<-listener
+			mgr.updateBackgrounds()
 			mgr.refreshBorders()
 			mgr.configureRoots()
 		}
```
Jacalz, line 154: Same as one of the comments above, if it works :)

Jacalz (changes requested):
Well done. The new images look great and are also tiny in size :)

Jacalz (changes requested):
Sorry for the wait. This is really close but I think you might want to update the desktop screenshots (in the project root) before merging.

internal/x11/wm/desk.go
Jacalz: Actually, I still think this hasn't been completed. Can be tidied up to `for range listener` while at it :)

andydotxyz:
> Sorry for the wait. This is really close but I think you might want to update the desktop screenshots (in the project root) before merging.

Good point thanks

internal/x11/wm/desk.go
andydotxyz: Yup, I missed that there were 4 occurrences over 2 files sorry

Jacalz (approved):
Well done :)

PR #239, opened 2023-01-07
Add a very basic virtual desktop module

### LINKED ISSUE #186: Support virtual desktops
Let's make better use of the space we have :)

Needs a pager too (separate issue at #28) and some handy keyboard shortcuts.

Should they be per-screen or refer to all-screens when shifting?

andydotxyz:
This is the bare bones. It is functionally complete but slow :(.

Mostly the slowness is due to unmapping/mapping all the windows.
I don't know how else to do it just now - once we have a built-in compositor we can do it super fast by just not drawing...

Fixes #186

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.
- [x] Public APIs match existing style.

vdesk-narrow
vdesk


andydotxyz:
No negative reviews, PR defaulting to accepted - it has been played with in the community

PR #241, opened 2023-01-09
Fix/lint issues

rafaelsousa:
### Description:

Fixes #(issue)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:

- [ ] Public APIs match existing style.
- [ ] Updated the vendor folder (using `go mod vendor`).

### DIFF
--- a/modules/systray/main.go
+++ b/modules/systray/main.go
@@ -206,10 +206,14 @@ func (t *tray) RegisterStatusNotifierItem(service string, sender dbus.Sender) (e
 func (t *tray) RegisterStatusNotifierHost(service string, sender dbus.Sender) (err *dbus.Error) {
 	log.Println("Register Host", service, sender)
 
-	watcher.Emit(t.conn, &watcher.StatusNotifierWatcher_StatusNotifierHostRegisteredSignal{
+	e := watcher.Emit(t.conn, &watcher.StatusNotifierWatcher_StatusNotifierHostRegisteredSignal{
 		Path: dbus.ObjectPath(service),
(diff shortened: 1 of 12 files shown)


andydotxyz (changes requested):
We can't use os.ReadFile yet as we support back to Go 1.14 and it was not present then.

Jacalz:
It looks like there still are a few places where ioutil has been removed. That will need to be added back.

rafaelsousa:
> We can't use os.ReadFile yet as we support back to Go 1.14 and it was not present then.

The os.Readfile references were reverted.

modules/systray/main.go
```diff
@@ -206,10 +206,14 @@ func (t *tray) RegisterStatusNotifierItem(service string, sender dbus.Sender) (e
 func (t *tray) RegisterStatusNotifierHost(service string, sender dbus.Sender) (err *dbus.Error) {
 	log.Println("Register Host", service, sender)
 
-	watcher.Emit(t.conn, &watcher.StatusNotifierWatcher_StatusNotifierHostRegisteredSignal{
+	e := watcher.Emit(t.conn, &watcher.StatusNotifierWatcher_StatusNotifierHostRegisteredSignal{
 		Path: dbus.ObjectPath(service),
 		Body: &watcher.StatusNotifierWatcher_StatusNotifierHostRegisteredSignalBody{},
 	})
+	// TODO: See the need of returning this error
```
andydotxyz, line 213: Is this TODO supposed to be there? Not sure I understand the comment

andydotxyz (commented):
Looking great thanks - just checking about the TODO

andydotxyz (approved):

PR #248, opened 2023-06-19
use go embed for resources in theme

taigrr:
### Description:
Remove fyne bundle dependency in fynedesk and switch to go:embed (w/go v1.17)

### Checklist:

- [x] Tests included.
- [x] Tests all pass.

#### Where applicable:
- [X] Updated the vendor folder (using `go mod vendor`).


taigrr:
Note that the 1.14 ubuntu/macos tests will probably fail, because the new tests are added in this PR, so both newly added tests and prior required tests are included in the test results. We can disregard these failures since the v1.17 and v1.20 tests succeeded

Jacalz (approved):
Looks good to me. I'm fine with it as long as @andydotxyz is too (and he'll needs to change the branch rules before it can be merged either way)

andydotxyz:
The rules and workflow should all be sorted on `develop` so I think a rebase should make it all pass.

PR #256, opened 2024-02-11
Add quake mode terminal module

andydotxyz:
Last feature addition for v0.4 I think :)

https://github.com/FyshOS/fynedesk/assets/294436/6dab63c2-0e55-4caa-a492-25b03318de47

### Checklist:

- [ ] Tests included. <- I couldn't figure out how to test the show, which is the only real code addition as the widget is from terminal package
- [x] Tests all pass.

### DIFF
--- a/modules/quaketerm/term.go
+++ b/modules/quaketerm/term.go
@@ -1,0 +1,164 @@
+package launcher
+
+import (
+	"time"
+
+	"fyshos.com/fynedesk"
+	"fyshos.com/fynedesk/internal/ui"
+	wmTheme "fyshos.com/fynedesk/theme"
+	"github.com/fyne-io/terminal"
+
+	"fyne.io/fyne/v2"
+	"fyne.io/fyne/v2/canvas"
+	"fyne.io/fyne/v2/container"
+	"fyne.io/fyne/v2/driver/desktop"
+	"fyne.io/fyne/v2/theme"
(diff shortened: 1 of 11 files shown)


modules/quaketerm/term.go
```diff
@@ -0,0 +1,164 @@
...
+	t.ui = win
+
+	go func() {
+		err := console.RunLocalShell()
+		if err != nil {
+			fyne.LogError("Failed to open terminal", err)
+		}
+		t.hide() // terminal exited
+
+		t.createTerm() // reset for next usage
```
taigrr, line 73: I have tmux auto-launch from my .bashrc (with a check to make sure it isn't already running.) The way this is written here will cause my tmux sessions to pile up rather quickly, if I'm understanding it correctly.

Looks like this behavior might not play nicely with pinentry-ncurses for Yubikey users. Is there a way to only spawn the shell JIT, right before the user actually sees it, rather than leaving it open in the background 24/7 if the module is enabled?

taigrr (commented):
slightly concerned about implications for sessionized tmux / pinentry but love the idea

andydotxyz:
I made it lazy load the terminal, that should fix the concern I think?

taigrr (approved):

taigrr:
Looks good now!

PR #265, opened 2024-10-09
Link to app editor Fysion if it is installed

andydotxyz:
If the source is available and the app is installed let it be edited from the bar.

This will be a killer feature for FyshOS release :)

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module fyshos.com/fynedesk
 go 1.19
 
 require (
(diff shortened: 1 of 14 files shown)


andydotxyz:
All right, this now gets the source from apps who's metadata says the source is available.
It clones the code and opens the editor :)

andydotxyz:
No comments so landing :)

PR #267, opened 2025-01-25
Embed a nice screensaver so we don't depend on XScreensaver

andydotxyz:
### Description:

Still needs:

- [x] Inhibit when requested by movie players etc
- [x] Multiple monitor support

### Checklist:

- [ ] Tests included.
- [x] Tests all pass.
- [x] Public APIs match existing style.


taigrr (commented):
Looks like we kinda have half-support for multimonitor through xscreensaver right? Just not if xscreensaver isn't installed?

andydotxyz:
Yes the xscreensaver plugin supports multi-monitor but the replacement with Fyne does not yet. Other than that it seems to be working quite well.

PR #279, opened 2026-01-20
Simplify (and fix on OpenBSD) ifconfig call

sdassow:
### Description:

Reduce number of commands by doing regex matching in Go to make it work on OpenBSD.

### Checklist:

- [ ] Tests included.
- [X] Lint and formatter run with no errors.
- [ ] Tests all pass.

#### Where applicable:


sdassow:
Ooops, this isn't right yet, sorry.

PR #280, opened 2026-01-20
Simplify ifconfig call and move regex logic to Go

sdassow:
### Description:

This should make network detection more reliable by reducing dependencies that aren't available everywhere out of the box (bash, pcregrep). And it adds support for IPv6.

### Checklist:

- [ ] Tests included.
- [X] Lint and formatter run with no errors.
- [ ] Tests all pass.

#### Where applicable:

### DIFF
--- a/modules/status/network.go
+++ b/modules/status/network.go
@@ -5,6 +5,7 @@ import (
 	"log"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


andydotxyz (approved):
Thanks :)

